PR #127, opened 2026-02-05
feat: support SDMX proxy #126

### LINKED ISSUE #126: Support integration with SDMX 3.0 proxy API
### StatGPT Backend version

0.4.0

We want to have access to SDMX 3.0 proxy for improved performance and specific statgpt analytics 

We should create additional sdmx client implementation to support proxy with sdmx 3.0

_No response_

kryachkow:
### Applicable issues

- #126 

### Description of changes

- Introduced new YAML configurations for IMF proxy channels, data sources
- Implemented a new database migration for the Proxy SDMX 3.0 data source type.
- Updated the data manager to support the new Proxy SDMX 3.0 configuration.
- Enhanced existing schemas to accommodate new proxy data structures.

This update enhances the functionality and data handling capabilities of the SDMX proxy system.

Checklist
 

- [ ]  Deployed and tested in a Review environment.


statgpt/common/data/sdmx/v21/sdmx_client.py
```diff
@@ -55,6 +83,30 @@ def __init__(
...
+    def _preview_body(body: Any, limit: int = 1000) -> str:
+        if body is None:
+            return ""
+        if isinstance(body, (bytes, bytearray)):
+            text = body[:limit].decode("utf-8", errors="replace")
+        else:
+            text = str(body)
+        if len(text) > limit:
+            return f"{text[:limit]}...(truncated)"
+        return text
```
Fedir-Yatsenko: I couldn't find any usages of using these functions: `_redact_headers` and `_preview_body`. I assume they were added for debugging purposes. So I guess we can delete them, right? Or please let me know if they are used.

pyproject.toml
```diff
@@ -181,6 +181,7 @@ aidial-rag-eval = ">=0.2.0,<0.3.0"
 # jinja2 is required for the aidial-rag-eval package, since it uses langchain with Jinja2 templates, which are optional for langchain
 jinja2 = "^3.1.6"
 types-aiofiles = "25.1.0.20251011"
+ipykernel = "^7.1.0"
```
Fedir-Yatsenko: Not sure if this dependency should be listed in this section.

statgpt/common/data/common/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    Operator,
+    Sdmx30AnnotationModel,
+    Sdmx30DataComponentFilter,
+    build_availability_filters,
+    to_content_constraint,
+    to_structure_message,
+)
+from .updated_at_mixin import UpdatedAtMixin
+
+__all__ = [
```
Fedir-Yatsenko: `__all__` doesn't add value here. Everything listed in it is already explicitly imported at the top of the file, so a `from statgpt.common.data.common import *` wildcard import would pick up exactly those names regardless — `__all__` just duplicates the import list.                                                         

`__all__` is useful when:
  - A module defines names locally and you want to control which ones are public
  - You want to exclude some imported names from wildcard exports
  - The module uses `import` statements that bring in names you don't want re-exported

None of those apply here. Consider removing `__all__`.

statgpt/common/data/base/property_source.py
```diff
@@ -0,0 +1,24 @@
+from enum import StrEnum
+
+from pydantic import Field
+
+from statgpt.common.schemas.base import BaseYamlModel
```
Fedir-Yatsenko: This import introduces a dependency from `data` to `schemas`, which historically we've kept separate — the `data` package is independent of `schemas` and `models`. Consider using `BaseModel` from `statgpt/common/data/base/base.py` instead to maintain that boundary.

statgpt/common/data/proxy/sdmx_schemas/structure_message.py
```diff
@@ -0,0 +1,162 @@
...
+    @classmethod
+    def get_from(
+        cls,
+        key: dict[str, list[str]] | None,
+        params: dict[str, str] | None,
+        dsd: DataStructureDefinition | None,
+    ) -> "ProxyAvailabilityRequestBody":
+        return cls(
+            filters=build_availability_filters(ProxyDataComponentFilter, key, params),
+            key=cls._build_key_segment(key=key, dsd=dsd),
```
Fedir-Yatsenko: Do we need a `key` if we provide `filters`? This seems redundant.

statgpt/common/data/proxy/sdmx_schemas/structure_message.py
```diff
@@ -0,0 +1,162 @@
...
+    data_constraints: list[ProxyDataConstraint] = Field(
+        alias="dataConstraints", default_factory=list
+    )
+
+
+class ProxyDataComponentFilter(Sdmx30DataComponentFilter):
+    pass
+
+
+class ProxyAvailabilityRequestBody(BaseModel):
```
Fedir-Yatsenko: This is a follow up to the [previous comment](#discussion_r2904052859). I would combine `ProxyAvailabilityRequestBody` and `QhAvailabilityRequestBody` into one class. Or there's a reason to keep them separate?

statgpt/common/data/proxy/sdmx_schemas/structure_message.py
Fedir-Yatsenko: I mean, the `filters` are always provided, so the `key` is simply duplicating the information in another form.

statgpt/common/data/common/__init__.py
Fedir-Yatsenko: statgpt/common/data/proxy/sdmx_schemas/__init__.py
statgpt/common/data/proxy/v30/__init__.py
statgpt/common/data/proxy/__init__.py

statgpt/common/data/base/sdmx_schemas/sdmx_plus.py
Fedir-Yatsenko, line 1: Since this is now part of the basic package, let's think about what is part of the SDMX standard, and what is SDMX+? For SDMX+ schemas, I would name the file something like `sdmx_plus_schemas.py`. POST is definitely not part of the standard, at least not yet, but I believe `Annotation` is, so we can leave it in this file.

Fedir-Yatsenko: Or there may be some hierarchy, such as:
```
sdmx_schemas/
├── __init__.py
├── sdmx_30.py
└── sdmx_plus.py
```

statgpt/common/data/proxy/v30/datasource.py
```diff
@@ -0,0 +1,241 @@
...
+        return QuanthubDataSetConfig.model_validate(d)
+
+    async def create_sdmx_client(self, auth_context: AuthContext) -> AsyncProxySdmxClient:
+        rate_limiter = await SdmxRateLimiterFactory.get(
+            self._config.get_id(), self._config.rate_limits
+        )
+        return AsyncProxySdmxClient.from_config(self._config, auth_context, rate_limiter)
+
+    def _should_use_cache(self, allow_cached: bool) -> bool:
+        auth_enabled = getattr(self._config, "auth_enabled", False)
```
Fedir-Yatsenko: There is no such property in the `ProxySdmx30DataSourceConfig` class, and I'm not sure if it will appear in the near future.

statgpt/common/data/statgpt_sdmx_proxy/config.py
```diff
@@ -0,0 +1,15 @@
+from pydantic import Field
+
+from statgpt.common.data.sdmx.common.config import SdmxDataSourceConfig
+
+
+class ProxySdmx30DataSourceConfig(SdmxDataSourceConfig):
+    """Configuration for SDMX 3.0 proxy data sources that still use sdmx1 parsing."""
+
+    versions: set[str] = Field(
```
Fedir-Yatsenko, line 9: Is this used anywhere?

statgpt/common/data/quanthub/v21/datasource.py
```diff
@@ -61,15 +63,76 @@ async def is_dataset_available(self, config: dict, auth_context: AuthContext) ->
...
-                    # 403 means user doesn't have access to dataset
                     return False
                 else:
-                    raise e
+                    raise
 
-    async def _get_dataset(
+    def _should_use_cache(self, allow_cached: bool) -> bool:
+        auth_enabled = getattr(self._config, "auth_enabled", False)
+        return allow_cached and not auth_enabled
```
Fedir-Yatsenko, line 75: Whenever possible, let's use the field directly rather than `getattr`. Or is there a reason why not to do the following way here?
```python
    def _should_use_cache(self, allow_cached: bool) -> bool:
        return allow_cached and not self._config.auth_enabled
```

statgpt/common/data/quanthub/v21/qh_sdmx_client.py
```diff
@@ -209,7 +209,7 @@ async def _qh_available_constraint(
             {'accept': 'application/json'}, Resource.availableconstraint
         )
         key = {} if key is None else key
-        req_body_obj = QhAvailabilityRequestBody.get_from(key=key, params=params)
+        req_body_obj = PostAvailabilityRequestBody.get_from(key=key, params=params)
```
Fedir-Yatsenko: Is `PostAvailabilityRequestBody` supposed to be used in the QuantHub package? If so, perhaps we should move this into the SDMX+ schemas of base package? And give it a more general name?

statgpt/common/data/statgpt_sdmx_proxy/v30/reader.py
Fedir-Yatsenko, line 1: There may be some code duplication from the `sdmx1` library, let's try to remove it.

kryachkow: duplication removed

statgpt/common/data/quanthub/v21/qh_sdmx_client.py
kryachkow: Moved to sdmx+ schema

statgpt/common/data/quanthub/v21/datasource.py
kryachkow: removed auth_enabled check for now

statgpt/common/data/statgpt_sdmx_proxy/config.py
kryachkow: yes it used in corresponding datasource and sdmx_client

statgpt/common/data/statgpt_sdmx_proxy/v30/sdmx_client.py
```diff
@@ -0,0 +1,319 @@
...
+        *,
+        agency_id: str,
+        resource_id: str,
+        version: str,
+        use_cache: bool = False,
+        key: dict[str, list[str]] | None = None,
+        params: dict[str, str] | None = None,
+        dsd: DataStructureDefinition | None = None,
+    ) -> StructureMessage:
+        async with self._rate_limiter.availability_limiter():
```
Fedir-Yatsenko, line 54: ### Double rate limiter acquisition in `AsyncProxySdmxClient` (potential deadlock)

**Files:** `statgpt/common/data/proxy/v30/sdmx_client.py`

The `availableconstraint()` method acquires the availability rate limiter at line 47:
```python
async with self._rate_limiter.availability_limiter():
    return await self._proxy_available_constraint(...)
```

Then `_proxy_available_constraint()` acquires it **again** at line 174:
```python
async with self._rate_limiter.availability_limiter():
    response = await self._perform_request(req)
```

The same pattern occurs for `data()` (line 68) -> `_proxy_data()` -> `_perform_get()` (line 311):
```python
async with limiter():
    response = await self._perform_request(req)
```

Since the rate limiter uses `asyncio.Semaphore`, which is **not reentrant**, each call acquires two permits instead of one. **This will deadlock** when the number of concurrent requests reaches the semaphore limit - all permits will be held by outer acquisitions waiting for inner acquisitions that can never proceed.

statgpt/common/data/proxy/v30/sdmx_client.py
```diff
@@ -0,0 +1,319 @@
...
+        *,
+        agency_id: str,
+        resource_id: str,
+        version: str,
+        use_cache: bool,
+        key: dict[str, list[str]] | None,
+        params: dict[str, str] | None,
+        dsd: DataStructureDefinition | None,
+    ) -> StructureMessage:
+        """Fetch available constraints from the QuantHub SDMX API."""
```
Fedir-Yatsenko: "StatGPT SDMX Proxy", not "QuantHub" 😉

statgpt/common/data/quanthub/v21/datasource.py
Fedir-Yatsenko: It looks like this is now dead code.

statgpt/common/data/base/property_source.py
```diff
@@ -0,0 +1,26 @@
...
+    CITATION = "citation"
+    """Retrieve the property from the dataset citation configuration."""
+    VALUE = "value"
+    """Use a fixed value specified in the `field`."""
+
+
+class PropertySource(BaseModel):
+    model_config = ConfigDict(
+        alias_generator=alias_generators.to_camel, populate_by_name=True, extra="allow"
+    )
```
Fedir-Yatsenko: I would use `BaseModel` from the `statgpt/common/data/base/base.py`  file

statgpt/common/data/statgpt_sdmx_proxy/v30/sdmx_client.py
kryachkow: removed accessing rate limiter on lower levels

statgpt/common/data/proxy/v30/sdmx_client.py
kryachkow: renamed

statgpt/common/data/base/property_source.py
kryachkow: make class use BaseModel

statgpt/common/data/sdmx/v21/sdmx_client.py
```diff
@@ -30,6 +30,14 @@ class AsyncSdmxClient:
 
     _LOADING: set[str] = set()
     """Urls of SDMX requests that are currently being loaded."""
+    _SENSITIVE_HEADER_KEYS = {
```
Fedir-Yatsenko: Is this used anywhere?

statgpt/common/data/quanthub/v21/dataset.py
Fedir-Yatsenko: Should we move this method to the `_truncate_fractional_seconds()`?

tests/integration/services/mocks.py
```diff
@@ -1,3 +1,4 @@
+# Mocks for integration tests. (dummy)
```
Fedir-Yatsenko: This comment was added but provides no value - the module name and location already convey this.

statgpt/common/data/quanthub/sdmx_schemas/v30/structure_message.py
```diff
@@ -1,87 +1,17 @@
...
+
+from statgpt.common.data.base.sdmx_schemas import (
+    Sdmx30AnnotationModel,
+    to_content_constraint,
+    to_structure_message,
+)
+
+
+class QhAnnotation(Sdmx30AnnotationModel):
+    pass
```
Fedir-Yatsenko: `Sdmx30AnnotationModel` is part of the SDMX standard, so I'm not sure if we need to create subclasses. I would use `Sdmx30AnnotationModel` directly.

Fedir-Yatsenko: (`ProxyAnnotation` as well)

statgpt/common/data/sdmx/v21/sdmx_client.py
```diff
@@ -87,7 +95,13 @@ async def categoryscheme(
...
 
     async def conceptscheme(
-        self, *, agency_id: str, resource_id: str, version: str, use_cache: bool = False
+        self,
+        *,
+        agency_id: str,
+        resource_id: str,
+        version: str,
+        use_cache: bool = False,
+        dsd_urn: str | None = None,
```
Fedir-Yatsenko: ### `dsd_urn` Parameter Added to Base Class But Unused

The `dsd_urn: str | None = None` parameter was added to `conceptscheme()`, `codelist()`, and `hierarchicalcodelist()` in the base `AsyncSdmxClient`, but the base implementations silently ignore it. Only the `AsyncProxySdmxClient` override passes it to `extra_headers`. This violates the principle of least surprise.

kryachkow: fixed

statgpt/common/data/quanthub/sdmx_schemas/v30/structure_message.py
kryachkow: fixed

tests/integration/services/mocks.py
kryachkow: removed

statgpt/common/data/quanthub/v21/dataset.py
kryachkow: moved to mixin

statgpt/common/data/sdmx/v21/sdmx_client.py
kryachkow: removed

kryachkow:
/deploy-review

>GitHub actions run: [24030498993](https://github.com/epam/statgpt-backend/actions/runs/24030498993)
>Environment URL: [review-environment](https://chat-statgpt-backend-pr-127.gke.test.dial.parts) | [pipeline](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/810878)

Fedir-Yatsenko (approved):

Issue #92, opened 2026-01-08
Core: Switch to GPT-5 series models


ypldan:
### StatGPT Backend version

0.2.0

### What is the problem this feature will solve?

GPT-4.1 will be deprecated in April 2026.

https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/model-retirements?view=foundry-classic&tabs=text#text-generation-1

### What is the proposed feature or solution?

We need to switch to GPT-5 series models and verify the changes.

navalnica:
- let's move to gpt-5.1 - it has 2026-11-13 mentioned as deprecation date
- (pitfall) gpt-5 models have changed set of parameters (temperature removed, thinking added). need to explore them and create appropriate configs for our usage

## Acceptance Criteria

- run evals before / after changes for all clients. compare metrics
- if there are regressions in LLM performance, tune prompts

## Additional

- try to move to gpt-5.2. it's a newer model, but it currently has no retirement date specified according to Azure document above

navalnica:
`gpt-5.2` was updated with following deprecation date: "No earlier than 2026-12-12"

navalnica:
plan:

1. let's migrate all chains to `gpt-5.2(reasoning none)`, except for `gpt-4.1-mini` (to `gpt-5-mini`)
1. run eval as is
1. tune prompts according to gpt-5 prompting guides and rerun evals

main questions we want to answer:

1. will migration with same prompts lead to regressions in quality and latency?
2. will migration lead to significant changes in hybrid indexing?
3. can we increase quality using gpt-5 models family by tuning prompts / model params?
4. can we use gpt-5 with reasoning for Supreme Agent?

PR #140, opened 2026-02-13
feat: add GPT-5 model family support #92

bahdan111:
### Applicable issues

- Connected to #92 (Haven't updated default models and configs)

### Description of changes

Adds GPT-5 model family support with conditional parameter handling:
- New models: `GPT_5_MINI`, `GPT_5_1`, `GPT_5_2`
- New enums: `ReasoningEffortEnum`, `VerbosityEnum`
- `LLMModelConfig` Pydantic validator enforces model-family constraints:
  - GPT-5: rejects `seed`, requires `reasoning_effort`, validates `temperature=1` when reasoning is enabled
  - Non-GPT-5: rejects `reasoning_effort` and `verbosity`
- Refactored `get_chat_model()` - moved params validation into `LLMModelConfig` 
- Removed unused `**kwargs` from `get_chat_model()` and `get_embeddings_model()`
- Moved `reasoning_effort` and `verbosity` defaults out of `LangChainSettings` into `LLMModelConfig`
- Default model remains GPT-4.1
- Fix bug with default embedding model incorrectly set in channel config, correct assignment: `llm_model: Mapped[str] = mapped_column(
        default=langchain_settings.embedding_default_model.value
    )`

### Checklist


bahdan111:
/deploy-review
>GitHub actions run: [21990045205](https://github.com/epam/statgpt-backend/actions/runs/21990045205)
>Environment URL: [review-environment](https://chat-statgpt-backend-pr-140.gke.test.dial.parts) | [pipeline](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/776490)

navalnica:
phrase "partially fixes #<issue number>" will likely close original issue. maybe rephrase?

statgpt/common/settings/langchain.py
```diff
@@ -35,11 +38,21 @@ class LangChainSettings(BaseSettings):
...
 
+    default_reasoning_effort: ReasoningEffortEnum | None = Field(
+        default=ReasoningEffortEnum.NONE,
+        description="Default reasoning effort for GPT-5 models (none/minimal/low/medium/high/xhigh)",
+    )
+
+    default_verbosity: VerbosityEnum | None = Field(
+        default=VerbosityEnum.LOW,
+        description="Default verbosity for GPT-5 models (low/medium/high). None means use model default.",
+    )
```
Fedir-Yatsenko: I would use `default=None` until `default_model` is from the GPT-5 family.

statgpt/common/utils/models.py
```diff
@@ -26,13 +27,24 @@ def get_chat_model(
...
+            params["reasoning_effort"] = model_config.reasoning_effort
+            if model_config.reasoning_effort == ReasoningEffortEnum.NONE:
+                params["temperature"] = 0
+        params.update({k: v for k, v in kwargs.items() if k not in ("temperature", "seed")})
+    else:
+        # Legacy models: use temperature and seed
+        params["temperature"] = model_config.temperature
+        if model_config.seed is not None:
+            params["seed"] = model_config.seed
+        params.update(kwargs)
```
Fedir-Yatsenko: Let's move this validation to the Pydantic model and remove `**kwargs` from this function. @navalnica, do you have any concerns about this suggestion?

statgpt/common/settings/langchain.py
navalnica: this MR should not change any defaults - this will be done in a separate MR

statgpt/common/settings/langchain.py
```diff
@@ -35,11 +38,21 @@ class LangChainSettings(BaseSettings):
...
-    default_seed: Optional[int] = Field(
+    default_seed: int | None = Field(
         default=None,
         description="Default seed for reproducible outputs",
     )
 
+    default_reasoning_effort: ReasoningEffortEnum | None = Field(
+        default=ReasoningEffortEnum.NONE,
+        description="Default reasoning effort for GPT-5 models (none/minimal/low/medium/high/xhigh)",
+    )
```
navalnica: same here. default must be None

statgpt/common/utils/models.py
navalnica: if it doesn't break function usages, let's remove kwargs

statgpt/common/utils/models.py
```diff
@@ -26,13 +27,24 @@ def get_chat_model(
...
         timeout=timeout,  # timeouts are crucial!
     )
-    params.update(kwargs)  # update default params
+
+    if model_config.deployment.is_gpt_5_family:
+        # GPT-5: use reasoning_effort, temperature only allowed with reasoning_effort=none
+        if model_config.reasoning_effort is not None:
+            params["reasoning_effort"] = model_config.reasoning_effort
+            if model_config.reasoning_effort == ReasoningEffortEnum.NONE:
+                params["temperature"] = 0
```
navalnica: @bahdan111 why not to set `params["temperature"]=model_config.temperature`?

statgpt/common/utils/models.py
navalnica: and let's move all checks to pydantic model. it's a good idea

statgpt/common/utils/models.py
Fedir-Yatsenko: Btw, if we move the validation into the model, we can do something like this:
```python

params.update(model_config.model_dump(exclude_none=True))
```

statgpt/common/settings/langchain.py
bahdan111: I would like to leave as is, since family 4 models would work even with this params set (we currently get rid of those during model validation anyway) While None don't work for 5 family models. Moreover, open ai uses higher reasoning and verbosity by default, which will add confusion if not  set explicitly

statgpt/common/utils/models.py
bahdan111: moved to pydantic model

statgpt/common/settings/langchain.py
bahdan111: answered in verbosity

statgpt/common/utils/models.py
bahdan111: refactored

bahdan111:
/deploy-review
>GitHub actions run: [22065546129](https://github.com/epam/statgpt-backend/actions/runs/22065546129)
>Environment URL: [review-environment](https://chat-statgpt-backend-pr-140.gke.test.dial.parts) | [pipeline](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/777611)

statgpt/common/settings/langchain.py
navalnica: let's move this field to `LLMModelConfig` class and include in model validation

navalnica: I see a separation of concerns issue between:
1. `LangChainSettings` and
2. `LLMModelConfig` / `EmbeddingsModelConfig`

let's do the following:

- delete all LLM setup from `LangChainSettings`
- move all LLM-related settings (including defaults) to `LLMModelConfig`:
  - deployment_id
  - temperature
  - seed (present only in Chat Completions API, marked as deprecated. absent in Embeddings and Responses API)
  - verbosity
  - thinking_effort
- make langchain settings responsible only for general langchain setup (and default embedding model name):
  - langchain's logging, verbosity
  - api version (NOTE: I think api version should be present only in langchain config, and never in LLM / Embedding configs. we, likely, wouldn't need to mix different API versions in the code for different models. so we can remove it from `statgpt.common.schemas.model_config.BaseModelConfig`)
  - default embedding model (NOTE: it's currently used in different places across the code. to avoid unnecessary refactoring, let's keep it in langchain config for now)

@Fedir-Yatsenko what do you think?

bahdan111:
/deploy-review
>GitHub actions run: [22066587191](https://github.com/epam/statgpt-backend/actions/runs/22066587191)
>Environment URL: [review-environment](https://chat-statgpt-backend-pr-140.gke.test.dial.parts) | [pipeline](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/777656)

navalnica:
there is a bug in `statgpt.common.common.models.Channel`:

- it specifies embedding model to be used by a channel, in `llm_model` field (field name is incorrect. and it likely led to the error below)
- the default value for `llm_model` field uses LLM model, not embedding model:

```py
llm_model: Mapped[str] = mapped_column(default=langchain_settings.default_model.value)
```

I suggest we use:

```py
llm_model: Mapped[str] = mapped_column(default=langchain_settings.embedding_default_model.value)
```

and delete default value in `EmbeddingsModelConfig`:

```py
class EmbeddingsModelConfig(BaseModelConfig):
    """Config for embeddings models."""

    deployment: EmbeddingModelsEnum = Field(
        description="The deployment of the model in DIAL"
    )
```

@Fedir-Yatsenko what do you think?

statgpt/common/utils/models.py
```diff
@@ -29,7 +29,7 @@ def get_chat_model(
         timeout=timeout,  # timeouts are crucial!
     )
 
-    params.update(model_config.model_dump(exclude_none=True))
+    params.update(model_config.model_dump(mode="json", exclude_none=True, exclude={"deployment", "api_version"}))
```
navalnica: 1. you can use `model_dump_json()`
2. no need to exclude "api_version" - we can delete `api_version=model_config.api_version,` instead

bahdan111: 1. we cannot use `model_dump_json()` since it returns string, and is not compatible with .update [here](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_dump_json)
2. removing "api_version" would require `get_embeddings_model` refactoring along with their pydantic models, we probably want to keep changes minimal?

statgpt/common/settings/langchain.py
Fedir-Yatsenko: > move all LLM-related settings (including defaults) to `LLMModelConfig`

The `LLMModelConfig` model cannot read values ​​from environment variables, so I would not change the existing behavior in this pull request. However, new fields can only be added to the `LLMModelConfig` class.

---

> I see a separation of concerns issue between:
> 
> 1. LangChainSettings and
> 2. LLMModelConfig / EmbeddingsModelConfig

Let's address this in a separate PR.

bahdan111:
/deploy-review
>GitHub actions run: [22094853247](https://github.com/epam/statgpt-backend/actions/runs/22094853247)
>Environment URL: [review-environment](https://chat-statgpt-backend-pr-140.gke.test.dial.parts) | [pipeline](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/778241)

statgpt/common/utils/models.py
Fedir-Yatsenko: Regarding the original comment, I have no objections, good suggestion, let's do it.

Fedir-Yatsenko: Regarding point 2, I agree with @navalnica that we can remove "api_version" from the `exclude` set. However, if the GPT-4.1 model works with `api-version`, I would not change this line: `api_version=model_config.api_version`, since we agreed that the default values ​​would be changed in a separate pull request.

statgpt/common/utils/models.py
Fedir-Yatsenko: Thanks!

statgpt/common/utils/models.py
Fedir-Yatsenko: Great!

statgpt/common/config/llm_models.py
```diff
@@ -6,6 +6,25 @@ class EmbeddingModelsEnum(StrEnum):
     TEXT_EMBEDDING_3_LARGE = "text-embedding-3-large"
 
 
+class ReasoningEffortEnum(StrEnum):
+    """Reasoning effort levels for GPT-5 models."""
+
+    NONE = "none"  # No reasoning mode - standard inference
```
Fedir-Yatsenko: Could you please use docstring instead of comment?

statgpt/common/schemas/model_config.py
```diff
@@ -30,16 +36,42 @@ class LLMModelConfig(BaseModelConfig):
...
+
+    @model_validator(mode="after")
+    def _normalize_model_family_params(self) -> "LLMModelConfig":
+        if self.deployment.is_gpt_5_family:
+            self.seed = None
+            if self.reasoning_effort == ReasoningEffortEnum.NONE:
+                self.temperature = 0
+            else:
+                # For reasoning levels only 1 is accepted
+                self.temperature = 1
```
Fedir-Yatsenko: Could you please throw a `ValidationError` exception instead of implicitly updating the value?

bahdan111:
> there is a bug in `statgpt.common.common.models.Channel`:
> 
> * it specifies embedding model to be used by a channel, in `llm_model` field (field name is incorrect. and it likely led to the error below)
> * the default value for `llm_model` field uses LLM model, not embedding model:
> 
> ```python
> llm_model: Mapped[str] = mapped_column(default=langchain_settings.default_model.value)
> ```
> 
> I suggest we use:
> 
> ```python
> llm_model: Mapped[str] = mapped_column(default=langchain_settings.embedding_default_model.value)
> ```
> 
> and delete default value in `EmbeddingsModelConfig`:
> 
> ```python
> class EmbeddingsModelConfig(BaseModelConfig):
>     """Config for embeddings models."""
> 
>     deployment: EmbeddingModelsEnum = Field(
>         description="The deployment of the model in DIAL"
>     )
> ```
> 
> @Fedir-Yatsenko what do you think?

shall we move to separate PR, I would also like to change misleading name "llm_model" and we would have to migrate db after. @navalnica

statgpt/common/utils/models.py
bahdan111: removed `api_version` from exclusion

statgpt/common/config/llm_models.py
bahdan111: done

Fedir-Yatsenko:
> shall we move to separate PR, I would also like to change misleading name "llm_model" and we would have to migrate db after. @navalnica

Let's first create an issue and then discuss it. If you want to change this, let's remove this column. Essentially, this parameter should be part of the Data Query config, not the channel.

statgpt/common/config/llm_models.py
Fedir-Yatsenko: Thank you!

statgpt/common/settings/langchain.py
bahdan111: changed model validation and removed default params from langchain setting

Fedir-Yatsenko: Great, thanks!

navalnica:
> > there is a bug in `statgpt.common.common.models.Channel`:
> > 
> > * it specifies embedding model to be used by a channel, in `llm_model` field (field name is incorrect. and it likely led to the error below)
> > * the default value for `llm_model` field uses LLM model, not embedding model:
> > 
> > ```python
> > llm_model: Mapped[str] = mapped_column(default=langchain_settings.default_model.value)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I suggest we use:
> > ```python
> > llm_model: Mapped[str] = mapped_column(default=langchain_settings.embedding_default_model.value)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and delete default value in `EmbeddingsModelConfig`:
> > ```python
> > class EmbeddingsModelConfig(BaseModelConfig):
> >     """Config for embeddings models."""
> > 
> >     deployment: EmbeddingModelsEnum = Field(
> >         description="The deployment of the model in DIAL"
> >     )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @Fedir-Yatsenko what do you think?
> 
> shall we move to separate PR, I would also like to change misleading name "llm_model" and we would have to migrate db after. @navalnica

let's change the varname in another PR

statgpt/common/settings/langchain.py
navalnica: > The LLMModelConfig model cannot read values ​​from environment variables

we likely don't need to control them with envvars anyway. it's more robust to use configs.

I agree to do it to another issue

bahdan111:
/deploy-review
>GitHub actions run: [22131897847](https://github.com/epam/statgpt-backend/actions/runs/22131897847)
>Environment URL: [review-environment](https://chat-statgpt-backend-pr-140.gke.test.dial.parts) | [pipeline](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/778912)

navalnica:
created follow-up issue to refactor langchain configs #150

statgpt/common/settings/langchain.py
Fedir-Yatsenko: #150

Fedir-Yatsenko (approved):

navalnica (approved):

PR #141, opened 2026-02-13
feat: add possibility to get dataset by urn without loading all datasets #142

### LINKED ISSUE #142: Add functionality to retrieve one specific dataset
### StatGPT Backend version

0.1.0

Current implementations like `get_dataset_by_source_id` or `list_available_dataset` loads all channel datasets which can be slow when we want to retrieve only one specific dataset.

Add `get_dataset_by_urn` method which will retrieve one specific dataset by urn specified

None alternatives, as all methods use implementation which loads all datasets

atomao:
### Applicable issues

- ~~fixes #142 (same info as here)~~
- ~~Current implementations like `get_dataset_by_source_id` or `list_available_dataset` loads all channel datasets which can be slow when we want to retrieve only one specific dataset.~~
- Above changes reverted and currently we added a possibiliy to get dataset/dataset versions models using dataset service (and then download dataset using handler)

### Description of changes
- ~~Added `get_dataset_by_urn` method to `ChannelServiceFacade` to retrieve only one dataset without loading all.~~
- Added `get_dataset_by_uuid` method to `DataSetService` to get specific dataset model by uuid.
- Added `get_last_completed_dataset_version_by_uuid` method to `DataSetService` to get specific dataset last completed version model by uuid.


navalnica:
could you pls create issue and link it here? this allows to include this PR/issue in release notes on the next release. do not forget to add issue number in the end of PR name. thanks

statgpt/app/services/chat_facade.py
```diff
@@ -411,6 +412,55 @@ async def _load_datasets(self, auth_context: AuthContext) -> list[VersionedDataS
...
+                    config=config_to_use,
+                    auth_context=auth_context,
+                    allow_offline=True,
+                    allow_cached=True,
+                )
+                if ds.status.status == 'online':
+                    res = VersionedDataSet(
+                        model=db_dataset, version=versions[db_dataset.id], data=ds
+                    )
+
```
navalnica: I'd rather first filter dataset models and then create VersionedDataSet instance:

```py
dataset_model = [ds for ds in dataset_models if UrnReference.model_validate(ds.details['urn']).short_urn() == dataset_urn.short_urn()]

if len(dataset_model) > 1:
  raise ValueError(...)

if not dataset_model:
  return None

model = dataset_model[0]

res = VersionedDataSet(model=model, ...)

# make other checks, like `status==online`

return res
```

statgpt/app/services/chat_facade.py
```diff
@@ -411,6 +412,55 @@ async def _load_datasets(self, auth_context: AuthContext) -> list[VersionedDataS
...
+                db_dataset_urn.short_urn() == dataset_urn.short_urn()
+            ) and await handler.is_dataset_available(config_to_use, auth_context):
+                ds = await handler.get_dataset(
+                    entity_id=db_dataset.id_,
+                    title=db_dataset.title,
+                    config=config_to_use,
+                    auth_context=auth_context,
+                    allow_offline=True,
+                    allow_cached=True,
+                )
```
navalnica: you still instantiate dataset instances in a loop. and you don't stop the loop once you found required dataset. this code does not introduce optimizations compared to the original one

statgpt/app/services/chat_facade.py
navalnica, line 465: I guess we now can optimize this function:

1. map source_id to dataset urn (filter dataset models from db)
2. instantiate dataset instance only once, using new `_get_dataset_by_urn` function

also, probably can optimize other functions in this module. @Fedir-Yatsenko , what do you think?

statgpt/app/services/chat_facade.py
atomao: done

statgpt/app/services/chat_facade.py
atomao: done, also renamed method from `_get_dataset...` to `get_dataset...`

statgpt/app/services/chat_facade.py
```diff
@@ -411,6 +412,61 @@ async def _load_datasets(self, auth_context: AuthContext) -> list[VersionedDataS
...
+        }
+        dataset_models = await dataset_service.get_datasets_models(
+            limit=None, offset=0, ids=versions.keys()
+        )
+        data_sources = {
+            ds.id: ds
+            for ds in await data_source_service.get_data_sources_models(
+                limit=None, offset=0, ids={ds.source_id for ds in dataset_models}
+            )
+        }
```
navalnica: I'd move this right before the usage of data_sources, since this code might not be even needed in case we raise or early return

navalnica: also, the code is confusing. it uses `ds` in two places with different meaning/value:

1. `for ds in dataset_models` - ds stands for a dataset model
2. `for ds in data_source_service.get_data_sources_models` - ds stands for datasource

pls fix

statgpt/app/services/chat_facade.py
```diff
@@ -411,6 +412,61 @@ async def _load_datasets(self, auth_context: AuthContext) -> list[VersionedDataS
...
+    ) -> VersionedDataSet | None:
+        dataset_service = DataSetService(self._session, session_lock=self._session_lock)
+        data_source_service = DataSourceService(self._session, session_lock=self._session_lock)
+
+        dataset_urn = UrnReference(agency_id=agency_id, resource_id=resource_id, version=version)
+        last_versions = await dataset_service.get_latest_successful_dataset_versions_for_channel(
+            channel_id=self._channel.id
+        )
+        versions = {
+            k: item.last_completed_version
```
navalnica: code style is odd. the standard is `key: value for key, value in dict.items()`. or `k: v`. not a critical, but a bit confusing

statgpt/app/services/chat_facade.py
```diff
@@ -411,6 +412,61 @@ async def _load_datasets(self, auth_context: AuthContext) -> list[VersionedDataS
...
+        }
+
+        dataset_models = [
+            ds
+            for ds in dataset_models
+            if UrnReference.model_validate(ds.details['urn']).short_urn() == dataset_urn.short_urn()
+        ]
+
+        if len(dataset_models) > 1:
+            raise ValueError(f"Multiple datasets found for the same URN: {dataset_urn.short_urn()}")
```
navalnica: @Fedir-Yatsenko , do you want to raise or simply pick the first model here?

statgpt/app/services/chat_facade.py
atomao: done

statgpt/app/services/chat_facade.py
atomao: done

statgpt/app/services/chat_facade.py
```diff
@@ -411,6 +412,66 @@ async def _load_datasets(self, auth_context: AuthContext) -> list[VersionedDataS
...
+
+        dataset_model = dataset_models[0]
+
+        data_sources = await data_source_service.get_data_sources_models(
+            limit=None, offset=0, ids={dataset_model.source_id}
+        )
+        if len(data_sources) > 1:
+            raise ValueError(
+                f"Multiple data sources found for the same dataset: {dataset_model.id}"
+            )
```
navalnica: @Fedir-Yatsenko same here: raise or pick first?

Fedir-Yatsenko: Honestly, I don't understand why we need to query for a list of items if we can retrieve an item by ID? (`get_by_id` or `get_schema_by_id`)

statgpt/app/services/chat_facade.py
Fedir-Yatsenko: I'm not sure that extracting a list of all datasets is the best option. Let's discuss.

statgpt/app/services/chat_facade.py
atomao: done for data sources, but for dataset models we first  need to fetch all versions and then fetch ds models to find model with correct urn

statgpt/common/services/dataset.py
```diff
@@ -222,6 +223,49 @@ async def get_model_by_id(self, item_id: int, expand: bool = False) -> models.Da
             await session.refresh(item, attribute_names=["source"])
         return item
 
+    async def get_dataset_by_uuid(
+        self, dataset_uuid: str | uuid.UUID, expand: bool = False
+    ) -> models.DataSet:
+        """Retrieve a models.DataSet by external UUID (`id_`) from dataset configs."""
+
+        query = select(models.DataSet).where(models.DataSet.id_ == dataset_uuid)
+        if expand:
```
navalnica, line 238: not sure both `str | uuid.UUID` will work here. could you test pls? @atomao

navalnica: claude code mentions that sqlalchemy will cast string to UUID internally, so string should be fine. the only argument to make explicit conversion with `uuid.UUID(value)` is to catch for malformed string

atomao: refactored it to explicitly convert str to UUID class object

PR #155, opened 2026-02-18
fix: gracefully handle hallucinated dataset IDs in populate_stage to prevent KeyError #152

### LINKED ISSUE #152: KeyError in populate_stage when LLM hallucinates dataset ID
### StatGPT Backend version

0.4.0

### Short description

During indicator selection, the LLM occasionally hallucinates a dataset UUID that doesn't exist in `datasets_dict`. The `populate_stage()` method (debug stage formatting) runs on the raw LLM response **before** `_remove_hallucinations()` gets a chance to filter them out. When `format_multidataset_queries()` tries to look up the hallucinated dataset ID, it crashes with a `KeyError`.

Run Data Query. Occasionally LLM will hallucinate dataset UUIDs (in my case it just misspelled one letter in UUID) - then this bug will happen

_No response_

### What do you see instead?

Error calling tool Query_Data: 
... 
File "pis_v1_v2.py", line 267, in async_lambda
    return await inputs[self.PARSED_RESPONSE_KEY].populate_stage(inputs)

File "pis_v1_v2.py", line 210, in populate_stage
    line = await formatter.format_multidataset_queries(queries=self.get_queries())

File "format_utils.py", line 112, in format_multidataset_queries
    dataset = self.datasets[dataset_id]

KeyError: 'UUID'

## Summary

During indicator selection, the LLM occasionally hallucinates a dataset UUID that doesn't exist in `datasets_dict`. The `populate_stage()` method (debug stage formatting) runs on the raw LLM response **before** `_remove_hallucinations()` gets a chance to filter them out. When `format_multidataset_queries()` tries to look up the hallucinated dataset ID, it crashes with a `KeyError`.

## Occurrence

Observed once during a full eval run. Rare — the LLM usually returns valid dataset IDs from the provided candidates, but occasionally fabricates or garbles a UUID.

## Root Cause

The chain execution order in `_create_chain_inner()` (`pis_v1_v2.py:269-276`):

```
format_candidates → LLM call → populate_stage() → _remove_hallucinations()
                                    ^ ERROR            ^ would have caught it
```

`populate_stage()` formats the raw LLM response for debug display and accesses `self.datasets[dataset_id]` (via `format_multidataset_queries`), which raises `KeyError` for a hallucinated dataset ID. The `_remove_hallucinations()` step, which is specifically designed to catch this, runs one step too late.

## Affected Files

- `statgpt/app/chains/data_query/query_builder/indicator_selection/format_utils.py:112` — `self.datasets[dataset_id]` lookup
- `statgpt/app/chains/data_query/query_builder/indicator_selection/semantic/packed_indicators_selection/pis_v1_v2.py:210,522,529,542` — `populate_stage` callers and dataset references section

### LLM Input (candidates presented to LLM)
https://epa.ms/jmQ9j3

### LLM Output (raw response with hallucinated ID)
https://epa.ms/Nd1VN5

veryniceuser:
Skip unknown dataset IDs with a warning log in format_multidataset_queries and V2 populate_stage dataset references, instead of crashing. This prevents a KeyError when the LLM hallucinates a dataset UUID before _remove_hallucinations gets a chance to filter it out.

### Applicable issues

- fixes #152 

### Checklist

### DIFF
--- a/statgpt/app/chains/data_query/query_builder/indicator_selection/format_utils.py
+++ b/statgpt/app/chains/data_query/query_builder/indicator_selection/format_utils.py
@@ -4,6 +4,7 @@ from statgpt.app.services.hybrid_searcher import HybridCandidateScored
 from statgpt.app.settings.dial_app import dial_app_settings
 from statgpt.app.utils.formatters import DatasetFormatterConfig, SimpleDatasetFormatter
 from statgpt.common.auth.auth_context import AuthContext
+from statgpt.common.config import logger
 from statgpt.common.data.base import CategoricalDimension, DataSet, DimensionQuery
 from statgpt.common.data.sdmx.v21.dataset import Sdmx21DataSet
 from statgpt.common.schemas.enums import LocaleEnum
@@ -109,7 +110,10 @@ class DatasetDimQueriesSimpleDictFormatter:
     ) -> str:
         lines = []
         for dataset_id, dataset_query in sorted(queries.items(), key=lambda x: x[0]):
-            dataset = self.datasets[dataset_id]
+            dataset = self.datasets.get(dataset_id)
+            if dataset is None:
+                logger.warning(f'Skipping unknown {dataset_id=} in query formatting')
+                continue
 
             prefix = '#' * header_level
             title = f'{prefix} {dataset.name}'
--- a/statgpt/app/chains/data_query/query_builder/indicator_selection/semantic/packed_indicators_selection/pis_v1_v2.py
+++ b/statgpt/app/chains/data_query/query_builder/indicator_selection/semantic/packed_indicators_selection/pis_v1_v2.py
@@ -541,7 +541,9 @@ it is especially true for dimension values like "all", "total", "all maturities"
                 for dataset_id in dataset_ids:
                     dataset = datasets_dict.get(dataset_id)
                     if dataset is None:
-                        logger.warning(f'Skipping unknown {dataset_id=} when populating stage content')
+                        logger.warning(
+                            f'Skipping unknown {dataset_id=} when populating stage content'
+                        )
                         continue
                     lines.append(f'* {dataset.name}')
                     lines.append(f'\t* ID: {dataset.source_id}')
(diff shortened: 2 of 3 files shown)


navalnica:
the issue is not only with formatting. if llm selected incorrect `uuid`, the whole query is unsable. we need to delete it inside indicator selection chain

navalnica:
also, it's better not to pass uuid to llm. instead, we can pass:
- dataset urn (imf:weo, for example)
- source_id/entity_id
- 1-based index

we also need to make sure there are no duplicates. for example if there are duplicated urns (for some reason. though they should not) we can add index prefix: "1-imf:weo"

veryniceuser:
> the issue is not only with formatting. if llm selected incorrect `uuid`, the whole query is unsable. we need to delete it inside indicator selection chain

I think it gets deleted in the latter stage at _remove_hallucinations()

veryniceuser:
> also, it's better not to pass uuid to llm. instead, we can pass:
> 
> * dataset urn (imf:weo, for example)
> * source_id/entity_id
> * 1-based index
> 
> we also need to make sure there are no duplicates. for example if there are duplicated urns (for some reason. though they should not) we can add index prefix: "1-imf:weo"

Sound like a bigger change. Should we merge this PR as a hotfix, and create a follow-up feature request to make these changes?

navalnica:
I don't like the current proposed change. it states that, when populating stage or formatting query, the query can contain dataset queries to unavailable datasets. this seems incorrect - we should remove any hallucinations from LLM response and call formatting only afterwards

veryniceuser:
> also, it's better not to pass uuid to llm. instead, we can pass:
> 
> * dataset urn (imf:weo, for example)
> * source_id/entity_id
> * 1-based index
> 
> we also need to make sure there are no duplicates. for example if there are duplicated urns (for some reason. though they should not) we can add index prefix: "1-imf:weo"

Created a separate feature request: #172, and pushed changes in PR #173. It's done to un-block this hotfix merge, and to have review and discussion for the in-depth fix in the separate place.

veryniceuser:
> I don't like the current proposed change. it states that, when populating stage or formatting query, the query can contain dataset queries to unavailable datasets. this seems incorrect - we should remove any hallucinations from LLM response and call formatting only afterwards

I disagree that proposed change is wrong, populate stage should not make implicit assumptions, and must not fail if the code that calls it will change in the future.

On the other hand, swapping remove_hallucinations and populate_stage makes sense. But because of the code structure, such change seem to require refactoring. And so should be prioritized separately

veryniceuser:
/deploy-review
>GitHub actions run: [22845383411](https://github.com/epam/statgpt-backend/actions/runs/22845383411)
>- **Deployment failed**: [failed](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/792023)

PR #173, opened 2026-02-27
feat: Add translation logic from UUID to Source ID and back in Indicator Selection Chain #172

### LINKED ISSUE #172: Avoid using Dataset UUIDs in prompt for indicator selection chain
### StatGPT Backend version

0.4.0

Dataset UUIDs generation by LLMs is prone to hallucinations. 

Dataset UUIDs come into indicator selection chain from along with indicator candidates: `inputs[self._candidates_key]` which we get by calling search_indicators_scored - it's tightly couple with our Data Layer. That's why indicator selection pipeline also uses Dataset UUID via datasets_dict everywhere.

We need to convert from Dataset UUIDs right before prompting the LLM, and then convert it back when processing the response.

Lets use DataSet.source_id - in SDMX datasets its more human readable than UUID in entity_id. E.g. 'imf:weo'

DataSet.source_id (dataset urn, imf:weo, for example)
1-based index

veryniceuser:
### Applicable issues

- fixes #172 

### Checklist

### DIFF
--- a/statgpt/app/chains/data_query/query_builder/indicator_selection/semantic/packed_indicators_selection/pis_v1_v2.py
+++ b/statgpt/app/chains/data_query/query_builder/indicator_selection/semantic/packed_indicators_selection/pis_v1_v2.py
@@ -69,10 +69,10 @@ class DatasetDimQueriesByRelevancy(BaseModel):
 class IndicatorCandidatesLLMFormatter:
     def __init__(
         self,
-        dataset_alias_to_name: dict[str, str],
+        dataset_id_to_name: dict[str, str],
         dataset_id_to_source_id: dict[str, str] | None = None,
     ):
-        self.dataset_alias_to_name = dataset_alias_to_name
+        self.dataset_id_to_name = dataset_id_to_name
         self._dataset_id_to_source_id = dataset_id_to_source_id or {}
 
     @classmethod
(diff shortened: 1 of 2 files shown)


statgpt/app/chains/data_query/query_builder/indicator_selection/semantic/packed_indicators_selection/pis_v1_v2.py
```diff
@@ -67,12 +67,19 @@ def combine_with_priority(self):
...
 
 class IndicatorCandidatesLLMFormatter:
-    def __init__(self, dataset_id_2_name: dict[str, str]):
-        self.dataset_id_2_name = dataset_id_2_name
+    def __init__(
+        self,
+        dataset_alias_to_name: dict[str, str],
+        dataset_id_to_source_id: dict[str, str] | None = None,
+    ):
+        self.dataset_alias_to_name = dataset_alias_to_name
```
navalnica: let's avoid using "alias" term - to avoid confusion with `dimension_alias` dataset config field

statgpt/app/chains/data_query/query_builder/indicator_selection/semantic/packed_indicators_selection/pis_v1_v2.py
```diff
@@ -129,6 +139,16 @@ class LLMResponseBase(BaseModel, ABC):
     def get_queries(self) -> DatasetDimQueriesType:
         pass
 
+    @abstractmethod
+    def translate_dataset_ids(self, source_id_to_dataset_id: dict[str, str]) -> None:
```
navalnica: let's use `translate_forward` and `translate_backward` for clarity

navalnica:
let's get back to this PR after we move to gpt-5* models - to simplify transition to gpt-5* (avoid unnecessary evals). we'll need to run following evals and compare:

- data query, before translation changes
- data query, after translation changes

to think: on full set of tests / subset, for one/multiple clients, with/without repro

navalnica (changes requested):

PR #184, opened 2026-03-05
fix: make RAG state model fields optional with defaults #189

### LINKED ISSUE #189: Bug: `!skip_tools_execution` returns incomplete FILE_RAG tool state
## StatGPT Backend version

 latest

## Short description 
`!skip_tools_execution` returns incomplete FILE_RAG tool state, causing eval parsing failures

## Steps to reproduce

1. Send a query with `!skip_tools_execution` command that triggers a `FILE_RAG` tool call
2. The backend returns `ToolMessageState(type=tool.tool_type)` with only `{'type': 'FILE_RAG'}`
3. Downstream consumers (e.g. eval framework) attempt to parse this as `DialRagState` and fail validation

Root cause: `supreme_agent.py` line ~120 — when `skip_tools_execution` is true, the tool response artifact uses the base `ToolMessageState` which only has `type`, but `DialRagState` requires `response`, `answered_by`, `pre_filter`, and `metadata`.

## Expected behavior

The skipped tool state should be valid against `DialRagState`. Either:
- Make `BaseRagState` and `DialRagState` fields optional (with defaults) so the minimal state validates
- Or return a properly populated `DialRagState` with empty/default values when skipping execution

## Actual behavior

```
pydantic_core._pydantic_core.ValidationError: 4 validation errors for StatGPTMultipleToolResponses
response
  Field required [type=missing, input_value={'type': 'FILE_RAG'}, input_type=dict]
answered_by
  Field required [type=missing, input_value={'type': 'FILE_RAG'}, input_type=dict]
pre_filter
  Field required [type=missing, input_value={'type': 'FILE_RAG'}, input_type=dict]
metadata
  Field required [type=missing, input_value={'type': 'FILE_RAG'}, input_type=dict]
```

## Additional information

Same issue likely applies to `DATA_QUERY` tool type — any tool with a specialized state subclass will fail when only the base `ToolMessageState` is returned during skip mode.

bahdan111:
### Applicable issues

- fixes #189 

### Description of changes

Update `BaseRagState` and `DialRagState` Pydantic models to make fields optional with sensible defaults:

- `BaseRagState`:  make `response` and `answered_by` optional (`str  = ""`)
- `DialRagState`: make `pre_filter` and `metadata` optional with `None` defaults
- `PreFilterResponse`: simplify field declarations by replacing `Field(default=None)` with plain `= None`

This allows partial construction of RAG state objects without requiring all fields upfront.

### Checklist

### DIFF
--- a/statgpt/app/schemas/file_rags/base.py
+++ b/statgpt/app/schemas/file_rags/base.py
@@ -4,7 +4,7 @@ from statgpt.common.schemas import RAGVersion, ToolTypes
 
 class BaseRagState(ToolMessageState):
     type: ToolTypes = ToolTypes.FILE_RAG
-    version: RAGVersion
+    version: RAGVersion = RAGVersion.DIAL
 
-    response: str  # This is not needed since we have content field
-    answered_by: str
+    response: str | None = None  # This is not needed since we have content field
+    answered_by: str | None = None
--- a/statgpt/app/schemas/file_rags/base.py
+++ b/statgpt/app/schemas/file_rags/base.py
@@ -4,7 +4,7 @@ from statgpt.common.schemas import RAGVersion, ToolTypes
 
 class BaseRagState(ToolMessageState):
     type: ToolTypes = ToolTypes.FILE_RAG
-    version: RAGVersion = RAGVersion.DIAL
+    version: RAGVersion
 
-    response: str | None = None  # This is not needed since we have content field
-    answered_by: str | None = None
+    response: str = ""  # This is not needed since we have content field
+    answered_by: str = ""
--- a/statgpt/app/schemas/file_rags/dial_rag.py
+++ b/statgpt/app/schemas/file_rags/dial_rag.py
@@ -292,18 +292,18 @@ class DialRagMetadata(BaseModel):
 
 
 class PreFilterResponse(BaseModel):
-    user_friendly_error: str | None = Field(default=None)
-    detailed_error: str | None = Field(default=None)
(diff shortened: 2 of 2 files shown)


statgpt/app/schemas/file_rags/base.py
```diff
@@ -4,7 +4,7 @@
 
 class BaseRagState(ToolMessageState):
     type: ToolTypes = ToolTypes.FILE_RAG
-    version: RAGVersion
+    version: RAGVersion = RAGVersion.DIAL
```
navalnica: let's avoid specifying version here. base class is not expected to be instantiated anywhere

statgpt/app/schemas/file_rags/base.py
```diff
@@ -4,7 +4,7 @@
...
 
-    response: str  # This is not needed since we have content field
-    answered_by: str
+    response: str | None = None  # This is not needed since we have content field
+    answered_by: str | None = None
```
navalnica: let's keep `str` type and use empty string as default

navalnica (changes requested):

statgpt/app/schemas/file_rags/base.py
bahdan111: done

statgpt/app/schemas/file_rags/base.py
bahdan111: done

navalnica (approved):

navalnica:
pls link to issue

PR #216, opened 2026-03-17
fix: database connection pool exhaustion during concurrent operations #174

### LINKED ISSUE #174: Database connection pool exhaustion during concurrent dataset operations
## StatGPT Backend version
0.2.0

## Short description
Database connection pool exhaustion (QueuePool limit reached) in admin dataset service on review environment.

## What steps will reproduce the bug?
1. Deploy the StatGPT backend to the review environment with default pool settings (`pool_size=20`, `max_overflow=10`).
2. Trigger multiple concurrent dataset operations (e.g., reloading/reindexing datasets, importing channels, or exporting data) through the admin API.
3. Background tasks spawned by these operations each acquire a database session and hold it for the entire duration of long-running indexing/export processes.
4. Once 30 connections are consumed (20 base + 10 overflow), any new request that needs a database connection will block for 30 seconds and then fail with the QueuePool timeout error.

## What is the expected behavior?
Database connections should be acquired and released promptly. Long-running background operations should not hold connections for extended periods, and the pool should have sufficient capacity for concurrent requests.

## What do you see instead?
```
ERROR:    | 2026-02-27 14:18:02 | 7 | statgpt.admin.services.dataset | QueuePool limit of size 20 overflow 10 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/20/3o7r)
```

All 30 available database connections are exhausted and new requests time out after 30 seconds.

## Additional information

**Environment**: Review environment.

### Pool Configuration

Defined in [database.py:78-84](statgpt/common/models/database.py#L78-L84):
```python
DEFAULT_ENGINE_CONFIG = dict(
    pool_size=20,
    max_overflow=10,
    pool_timeout=30,
    pool_recycle=1800,
    pool_pre_ping=True,
)
```

The `SessionMakerSingleton` uses a single engine instance, meaning this pool is shared across all request handlers and background tasks within the same process.

Fedir-Yatsenko:
### Applicable issues

- first part of the #174 

### Description of changes

Resolves database connection pool exhaustion (`QueuePool limit of size 20 overflow 10 reached`) caused by FastAPI's `Depends(get_session)` keeping sessions open for the entire lifetime of background tasks.

**Root cause:** When a FastAPI endpoint uses `Depends(get_session)` and schedules `BackgroundTasks`, the dependency-injected session is not closed until _all_ background tasks finish.

**Improvement:** `PgVectorStore` and admin services now manage their own short-lived sessions instead of holding a single session for the entire duration of long-running background operations (reindexing, import/export).

**Other changes:**

- **Return 400 for `InvalidConfigurationError` in reload-indicators endpoints** — both single-dataset and all-datasets reload endpoints now catch `InvalidConfigurationError` and return HTTP 400 instead of 500
- **Fix logger name for database log level setting** — corrected `"common.models.database"` → `"statgpt.common.models.database"` so the `level_db` setting actually applies

### Checklist

### DIFF
--- a/statgpt/admin/services/channel.py
+++ b/statgpt/admin/services/channel.py
@@ -18,6 +18,7 @@ from statgpt.common import utils
 from statgpt.common.auth.auth_context import AuthContext
 from statgpt.common.schemas import AuditActionType, AuditEntityType
 from statgpt.common.services import ChannelSerializer, ChannelService
+from statgpt.common.services.base import DbServiceBase
 from statgpt.common.settings.dial import dial_settings
 from statgpt.common.utils import dial_core_factory
 from statgpt.common.utils.elastic import ElasticSearchFactory
@@ -28,7 +29,9 @@ from .background_tasks import background_task
 _log = logging.getLogger(__name__)
 
 
-class AdminPortalChannelService(ChannelService):
+class AdminPortalChannelService(ChannelService, DbServiceBase):
+    def __init__(self, session=None):
+        DbServiceBase.__init__(self, session)
 
     @staticmethod
     def _parse_integrity_error(
@@ -125,7 +128,7 @@ class AdminPortalChannelService(ChannelService):
         return deleted_item
 
     async def _clear_vector_store(self, channel: models.Channel, auth_context: AuthContext) -> None:
-        vector_store_factory = VectorStoreFactory(session=self._session)
+        vector_store_factory = VectorStoreFactory()
 
         collections = [
             channel.indicator_table_name,
@@ -276,22 +279,28 @@ class AdminPortalChannelService(ChannelService):
         Deduplication is performed based on document content.
         Documents with identical content will be merged.
         """
-        channel = await self.get_model_by_id(channel_id)
-        vector_store_factory = VectorStoreFactory(session=self._session)
+        async with self._scoped_session():
+            channel = await self.get_model_by_id(channel_id)
+            # Extract scalar values while still inside the session scope:
+            available_dims_table = channel.available_dimensions_table_name
+            special_dims_table = channel.special_dimensions_table_name
+            llm_model = channel.llm_model
+
+        vector_store_factory = VectorStoreFactory()
 
         _log.info(f"Deduplicating available_dimensions for channel {channel_id}")
         available_dims_store = await vector_store_factory.get_vector_store(
-            collection_name=channel.available_dimensions_table_name,
+            collection_name=available_dims_table,
             auth_context=auth_context,
-            embedding_model_name=channel.llm_model,
+            embedding_model_name=llm_model,
         )
         await available_dims_store.deduplicate_by_document_content()
 
         _log.info(f"Deduplicating special_dimensions for channel {channel_id}")
         special_dims_store = await vector_store_factory.get_vector_store(
-            collection_name=channel.special_dimensions_table_name,
+            collection_name=special_dims_table,
             auth_context=auth_context,
-            embedding_model_name=channel.llm_model,
+            embedding_model_name=llm_model,
         )
         await special_dims_store.deduplicate_by_document_content()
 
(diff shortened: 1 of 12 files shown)


Fedir-Yatsenko:
/deploy-review
>GitHub actions run: [23202893955](https://github.com/epam/statgpt-backend/actions/runs/23202893955)
>Environment URL: [review-environment](https://chat-statgpt-backend-pr-216.gke.test.dial.parts) | [pipeline](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/798317)

statgpt/admin/services/channel.py
```diff
@@ -28,7 +29,9 @@
 _log = logging.getLogger(__name__)
 
 
-class AdminPortalChannelService(ChannelService):
+class AdminPortalChannelService(ChannelService, DbServiceBase):
+    def __init__(self, session=None):
+        DbServiceBase.__init__(self, session)
```
ypldan: should there be `ChanelService.__init__` here as well?

statgpt/common/services/base.py
```diff
@@ -1,19 +1,60 @@
...
+
     @asynccontextmanager
-    async def _lock_session(self):
+    async def _lock_session(self) -> AsyncIterator[AsyncSession]:
         """Acquire lock and yield session for thread-safe operations."""
         async with self._session_lock:
             yield self._session
+
+    @asynccontextmanager
+    async def _scoped_session(self) -> AsyncIterator[AsyncSession]:
```
ypldan, line 44: This method is definitely not thread safe with current implementation: if two coroutines call it, there could be a situation when session will be shared by them, but first coroutine can close it earlier than second one is no longer used.

I think it's worth putting some additional guardrail here.

statgpt/admin/services/dataset.py
```diff
@@ -2146,6 +2187,17 @@ def _format_result_summary(jobs: list[models.AutoUpdateJob]) -> str:
             parts.append(part)
         return ", ".join(parts)
 
+    async def _get_auto_update_job_or_raise(self, auto_update_job_id: int) -> models.AutoUpdateJob:
+        job: models.AutoUpdateJob | None = await self._session.get(
+            models.AutoUpdateJob, auto_update_job_id
+        )
+        if job is None:
+            raise HTTPException(
```
ypldan: As far as In understand that is HTTPException raised in background task. Looks strange to me.

ypldan (changes requested):

ypldan:
@Fedir-Yatsenko, one thought that I have after this review: should we switch to short-living sessions for chat completion (and user-faced MCP in the future)?

Fedir-Yatsenko:
> @Fedir-Yatsenko, one thought that I have after this review: should we switch to short-living sessions for chat completion (and user-faced MCP in the future)?

Definitely yes, I'm going to do that in the chat completion after this PR. That's why I'm not closing the issue with this PR, but just writing "first part of #174".

statgpt/admin/services/channel.py
Fedir-Yatsenko: Thanks for pointing this out! I missed this, in my opinion, this is the wrong hierarchy. The `ChannelService` should be a child class of the `DbServiceBase`, not of the `AdminPortalChannelService`.

Fedir-Yatsenko:
/deploy-review
>GitHub actions run: [23238468294](https://github.com/epam/statgpt-backend/actions/runs/23238468294)
>Environment URL: [review-environment](https://chat-statgpt-backend-pr-216.gke.test.dial.parts) | [pipeline](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/798770)

statgpt/admin/services/dataset.py
Fedir-Yatsenko: Fixed

statgpt/common/services/base.py
Fedir-Yatsenko: Agree, added fix. But it seems to me we need to rethink the architecture: I would move the interaction with the database from services to a separate package.

ypldan (approved):

PR #222, opened 2026-03-20
fix: remove DbServiceBase from PgVectorStore to allow concurrent session access #174

Fedir-Yatsenko:
### Applicable issues

- Second part of the #174

### Description of changes

During channel reindex, the hybrid indexer runs `chain.abatch(max_concurrency=N)` which triggers multiple concurrent calls to `PgVectorStore.search_with_similarity_score()` on the same instance. `DbServiceBase._scoped_session()` stores the session on the instance and uses a boolean flag that raises `RuntimeError` on concurrent access, causing the reindex to fail with an `ExceptionGroup`.

<details>

<summary>Error traceback</summary>

```
ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | RuntimeError: Concurrent _scoped_session() calls on the same instance are not supported.
    |   Create a separate service instance per concurrent coroutine, or provide a session at construction time.
    +------------------------------------
```

</details>

**Root cause:** `PgVectorStore` inherited `DbServiceBase` but used it in a degenerate way — always constructed with `session=None`, never using `_lock_session()` or `_session_lock`. The shared instance-level session storage was incompatible with concurrent coroutines.

**Fix:** Remove `DbServiceBase` from `PgVectorStore` and manage sessions directly:
- Read-only methods use `get_readonly_session_context_manager()` — each call gets an independent session
- Write methods use `get_session_context_manager()`
- Helper methods (`_create_table_if_not_exist`, `_acquire_dataset_lock`, `_import_documents_from_zipfile`, etc.) now accept an explicit `session` parameter instead of relying on `self._session`

This makes `PgVectorStore` inherently concurrency-safe — no shared mutable state between coroutines.

### Checklist

### DIFF
--- a/statgpt/common/vectorstore/pg_vector_store/pg_vector_store.py
+++ b/statgpt/common/vectorstore/pg_vector_store/pg_vector_store.py
@@ -22,7 +22,7 @@ from sqlalchemy.sql import text
 from sqlalchemy.sql.elements import TextClause
 from sqlalchemy.sql.expression import func
 
-from statgpt.common.services.base import DbServiceBase
+from statgpt.common.models import get_readonly_session_context_manager, get_session_context_manager
 from statgpt.common.settings.database import PostgresSettings
 from statgpt.common.settings.document import DimensionValueDocumentMetadataFields
 from statgpt.common.utils import batched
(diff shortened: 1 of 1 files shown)


Fedir-Yatsenko:
/deploy-review
>GitHub actions run: [23344749083](https://github.com/epam/statgpt-backend/actions/runs/23344749083)
>Environment URL: [review-environment](https://chat-statgpt-backend-pr-222.gke.test.dial.parts) | [pipeline](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/800805)

ypldan (approved):

PR #296, opened 2026-04-14
feat: add endpoint to generate Python code from JSON queries and implement related schemas #237

### LINKED ISSUE #237: Add python attachment creation endpoint
### StatGPT Backend version

latest

We want to have possibility for user changed queries to be parsed and returned as attachment 

Add provider field to query json and receive array of json queries to generate python attachment

_No response_

kryachkow:
### Applicable issues

Close #237 

### Description of changes

- Introduced a new POST endpoint `/python-attachment` to generate Python code based on provided JSON query objects.
- Added `GeneratePythonCodeRequest` and `GeneratePythonCodeResponse` schemas to handle request and response formats.
- Implemented the `generate_merged_python_code` function to compile Python code from multiple queries.
- Updated existing schemas and services to integrate the new functionality.

### Checklist


statgpt/common/data/sdmx/v21/dataset.py
Fedir-Yatsenko: Is it possible to reuse the `generate_python_query_body` function from the `statgpt/app/services/python_code_generator.py` module here?

Fedir-Yatsenko: Btw, maybe we should remove `get_python_code_body` from the dataset classes? And use `generate_merged_python_code(queries: list[JsonQueryWithMetadata])` instead.

statgpt/common/schemas/query.py
```diff
@@ -23,6 +23,13 @@ class JsonComponentQuery(BaseYamlModel):
     operator: JsonQueryOperator = Field(description="The operator of the query")
     values: list[str] = Field(description="The values of the query")
 
+    @field_validator("values", mode="before")
+    @classmethod
+    def _coerce_values(cls, v: object) -> list[str]:
+        if isinstance(v, str):
+            return v.split(",")
```
Fedir-Yatsenko: Could you please mention this behavior in the field description above?

statgpt/app/services/onboarding_response_appenders.py
```diff
@@ -67,6 +67,9 @@ async def _append_json_query_attachment(
         if dataset is None:
             _log.warning("Dataset with id %s not found", urn)
             return
+        sdmx1_source = getattr(dataset, 'resolved_sdmx1_source', None) or getattr(
+            dataset.config, 'sdmx1_source', None
```
Fedir-Yatsenko: Let's avoid usage of `getattr` when possible. In this case, the`resolved_sdmx1_source` abstract method can be added to the base class.

Btw, do we need second part: `getattr(dataset.config, 'sdmx1_source', None)`?

statgpt/app/services/python_code_generator.py
```diff
@@ -0,0 +1,118 @@
...
+            params["startPeriod"] = f.values[0]
+            params["endPeriod"] = f.values[1]
+        elif f.operator == JsonQueryOperator.GE and f.values:
+            params["startPeriod"] = f.values[0]
+        elif f.operator == JsonQueryOperator.LE and f.values:
+            params["endPeriod"] = f.values[0]
+    return params
+
+
+def _detect_time_component(filters: list[JsonComponentQuery]) -> str:
```
Fedir-Yatsenko: Let's add a new field for the time period dimension to the `JsonQueryMetadata` class and remove this detection.

statgpt/app/services/python_code_generator.py
```diff
@@ -0,0 +1,118 @@
...
+import sdmx"""
+
+_URN_PATTERN = re.compile(r"^(?P<agency>[^:]+):(?P<resource>[^(]+)\((?P<version>[^)]+)\)$")
+
+
+def _parse_urn(urn: str) -> tuple[str, str, str]:
+    """Parse short URN like 'AGENCY:RESOURCE(VERSION)' into (agency_id, resource_id, version)."""
+    match = _URN_PATTERN.match(urn)
+    if not match:
+        raise ValueError(f"Invalid URN format: {urn!r}")
```
Fedir-Yatsenko: If a malformed URN is passed via the API, `_parse_urn` raises a `ValueError` that is not caught in the endpoint handler. This will result in an HTTP 500 rather than a 422 validation error.

Let's add a `field_validator` on `JsonQuery.urn` to validate the URN format at deserialization time.

statgpt/common/data/sdmx/v21/dataset.py
kryachkow: moved common logic in `common `module, `generate_python_query_body` is placed in `app `layer when dataset in `common`

statgpt/common/schemas/query.py
kryachkow: mentioned in behaviour

statgpt/app/services/onboarding_response_appenders.py
kryachkow: provided abstract method and removed getattr

statgpt/app/services/python_code_generator.py
kryachkow: new time period dimension is added

statgpt/app/services/python_code_generator.py
kryachkow: validation added

statgpt/common/data/sdmx/v21/dataset.py
Fedir-Yatsenko: Great, thanks!

statgpt/common/schemas/query.py
```diff
@@ -21,23 +22,56 @@ class JsonQueryOperator(StrEnum):
...
+        description="The time period dimension code",
+    )
     dataset_url: str | None = Field(default=None, description="URL of the dataset")
 
 
 class JsonQuery(BaseYamlModel):
     urn: str = Field(description="The urn of the dataset")
     filters: list[JsonComponentQuery] = Field(description="The list of component queries")
 
+    _URN_PATTERN = re.compile(r"^(?P<agency>[^:]+):(?P<resource>[^(]+)\((?P<version>[^)]+)\)$")
```
Fedir-Yatsenko: ❌ `_URN_PATTERN` is a `ModelPrivateAttr`, not a regex — the validator *always* raises `AttributeError`:

```
AttributeError: 'ModelPrivateAttr' object has no attribute 'match'
```

In Pydantic v2, a class-level attribute with a leading underscore is automatically converted into a `ModelPrivateAttr` descriptor (unless it is annotated `ClassVar` or declared via `PrivateAttr`). Inside the `@field_validator`, `cls._URN_PATTERN.match(value)` therefore calls `match` on the descriptor object, not the compiled regex.

statgpt/common/data/base/dataset.py
```diff
@@ -238,6 +238,18 @@ def get_resolved_config(self) -> dict | None:
     def status(self) -> Status:
         pass
 
+    @abstractmethod
+    def get_resolved_sdmx1_source(self) -> str | None:
+        """Return resolved SDMX 1.0 source identifier if available."""
```
Fedir-Yatsenko: "SDMX 1.0" sounds like a reference to a standard, but the comment should point to the `sdmx1` library.

statgpt/app/services/onboarding_response_appenders.py
```diff
@@ -67,14 +67,20 @@ async def _append_json_query_attachment(
...
+            rest_key_dimension_codes=dataset.sdmx_rest_key_dimension_codes(),
         )
         json_query = JsonQueryWithMetadata.from_query(
-            query=self._get_relative_time_period_aware_query(self._response.query),
-            metadata=json_query_metadata,
+            query=self._get_relative_time_period_aware_query(
+                self._response.query, time_period_dimension
+            ),
+            sdmx1_source=sdmx1_source,
         ).model_dump(by_alias=True)
```
Fedir-Yatsenko, line 83: `from_query`'s signature (`statgpt/common/schemas/query.py:100-111`) declares `metadata: JsonQueryMetadata` **without a default**, so this call will raise `TypeError: from_query() missing 1 required keyword-only argument: 'metadata'` whenever the onboarding path fires a `PredefinedDataQueryResponse`.

The `json_query_metadata` object is built three lines above but never passed.

statgpt/common/schemas/query.py
kryachkow: fixed

Fedir-Yatsenko:
/deploy-review
>GitHub actions run: [24777239852](https://github.com/epam/statgpt-backend/actions/runs/24777239852)
>Environment URL: [review-environment](https://chat-statgpt-backend-pr-296.gke.test.dial.parts) | [pipeline](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/821153)

Fedir-Yatsenko (approved):

PR #351, opened 2026-05-06
feat: [app] add interceptable command to skip guardrails #346

### LINKED ISSUE #346: [app] add interceptable command to skip guardrails
### StatGPT Backend version
latest (development)

We run tool-selection evals against the backend to measure whether the supreme agent picks the correct tool(s) for each test-case message. The out-of-scope guardrail (statgpt/app/chains/out_of_scope_checker.py) sits **upstream** of the
agent. When that happens:

  - The supreme agent never runs.
  - `response.tool_calls` is empty.
  - The tool-selection checker returns `not_applicable` for that row.

The result is a variable, run-dependent number of "not applicable" rows that masks the true tool-selection pass-rate and makes run-over-run comparisons unreliable.

The existing `SKIP_OUT_OF_SCOPE_CHECK` env var is deploy-time only — it can't be toggled per request, so it can't be turned on for an eval run without mutating shared backend config.

`!out_of_scope_only` does **not** solve this either: it is designed to test the guardrail itself, and its eval flag (EVAL_CMD_OUT_OF_SCOPE_ONLY) also disables tool-selection scoring, normalization checks, and RAG checks — the
opposite of what a tool-selection eval needs.

Add a per-request guardrail-skip mechanism that:

  - Causes the OOS checker to write `OUT_OF_SCOPE=None` (NOT `False`) — the  sentinel that means "guardrail did not run", so the eval marks those rows `is_applicable=False` rather than treating them as passes.
  - Lets the supreme agent run normally so tool-selection metrics reflect real model behavior.

**Implementation: `!skip_guardrails` interceptable command**:
  - Add state var key: `StateVarsConfig.CMD_SKIP_GUARDRAILS` 
  - Add settings field: `CMD_SKIP_GUARDRAILS` (default false)
  - Register the command
  - Initialize state default
  - Add skip logic in the pipeline
  - Wire new command with the evaluation pipeline

**Acceptance criteria:**
- [ ] Backend command registered, default off; no behavior change in prod
- [ ] OOS checker writes `OUT_OF_SCOPE=None` when flag is set (verified: not
      `False`, so the eval correctly marks the row "not applicable" rather
      than counting it as a guardrail pass)
- [ ] Supreme agent runs normally; tool-selection eval produces real
      pass/fail values (no longer `not_applicable` due to OOS misfire)

**Alternative approach — `skip_guardrails` field on `StatGPTConfiguration`**.
Carry the skip as a structured field in `request.custom_fields.configuration`
instead of an `!` prefix in the message text. Requires:
  - New `skip_guardrails: bool` field on `StatGPTConfiguration`
    (statgpt/app/schemas/dial_app_configuration.py) — `extra='allow'` already
    set, so existing callers are unaffected
  - Same single check in `out_of_scope_checker._stream_response()`, reading
    via `ChainParameters.get_configuration(inputs)`
  - Eval runner sends it as a JSON body field instead of injecting a `!`
    prefix into messages

| | `!skip_guardrails`  | `skip_guardrails` config field |
|---|---|---|
| Control signal location | Inside message text (stripped before LLM) | Request metadata, never near prompts |
| Gated by `ENABLE_DEV_COMMANDS` | **Yes** (auto, by interceptor) | No — must be implemented separately |
| Fail-fast on misconfigured backend | Yes  | Must be added separately |
| Consistent with existing patterns | Yes — joins `!out_of_scope_only`, `!skip_tools_execution`, etc. | New mechanism for this purpose |
| Works from any DIAL client | Any client that can send a `!` prefix | Any client supporting `custom_fields.configuration` |

**Why we chose the command approach:** the `ENABLE_DEV_COMMANDS` gate is a hard production safety boundary, and the interceptor pattern enforces it automatically — matching every other dev-only flag in the codebase. The config-field approach is cleaner architecturally (no string parsing of message bodies) but loses a guard we get for free with the command path.

Azamat0315277:
### Applicable issues

- feat #346 

### Description of changes

Renames `skip_out_of_scope_check` to `cmd_skip_guardrails` and migrates it from a static chain input to an interceptable state variable, consistent with the existing `CMD_*` command pattern.

  **Changes:**
  - Rename env var `SKIP_OUT_OF_SCOPE_CHECK` → `CMD_SKIP_GUARDRAILS` in `DialAppSettings`
  - Remove `ChainParameters.skip_out_of_scope_check()` helper and its `SKIP_OUT_OF_SCOPE_CHECK` key from `ChainParametersConfig`
  - Add `CMD_SKIP_GUARDRAILS` to `StateVarsConfig` and populate it in `init_state` (alongside other `CMD_*` flags)
  - `OutOfScopeChecker` now reads the flag from state instead of chain inputs
  - Register `skip_guardrails` as an `InterceptableCommand` in `CommandsInterceptor`

### Checklist

### Status

Draft


Azamat0315277:
/deploy-review
>GitHub actions run: [25432347938](https://github.com/epam/statgpt-backend/actions/runs/25432347938)
>Environment URL: [review-environment](https://chat-statgpt-backend-pr-351.gke.test.dial.parts) | [pipeline](https://gitlab.deltixhub.com/Clients/dial/test/statgpt-review/-/pipelines/829644)

navalnica:
seems good now. I want to ensure `ENABLE_DEV_COMMANDS` works as expected. will approve once I verify it myself

PR #378, opened 2026-05-19
chore: update dependencies and refactor DIAL API integration (#373)

### LINKED ISSUE #373: Refactoring: switch to `aidial-client` for interactions with DIAL API
### StatGPT Backend version

latest

Currently we have different implementation for DIAL Client, and there's no one unified approach.

[aidial-client](https://github.com/epam/ai-dial-client-python) implements DIAL client logic in python. Let's reuse it instead of our local implementations.

_No response_

kryachkow:
## Applicable issues

#373

## Summary

Replaces the hand-rolled `DialCore` httpx wrapper with the official [`aidial-client`](https://github.com/epam/ai-dial-client-python) SDK (`AsyncDial`) for all DIAL API interactions — file storage, model pricing, and user-info/role resolution. This removes ~250 lines of bespoke HTTP plumbing and delegates connection lifecycle, retries, and response parsing to the maintained SDK.

### New DIAL client module (`statgpt/common/utils/dial/client.py`)
- `dial_client_factory` — async-context factory yielding an `AsyncDial` with deterministic pool cleanup via the SDK's `aclose()`.
- `open_file_stream` / `write_file_to` — streaming downloads that never buffer full files in memory (safe for large exported channel archives).

### Attachments (`attachments.py`)
- `AttachmentsStorage` now wraps `AsyncDial` directly; listing uses the SDK's `limit`/`token` pagination.
- Drops the local `AttachmentResponse` model in favor of the SDK's `FileItem`.
- Adds `_ProgressBufferedReader` (a real `BufferedReader` subclass) to satisfy the SDK's pydantic-v1 upload validation while streaming uploads from disk with progress logging; resets progress accounting on reader rewind (SDK retries).

### Model pricing & token cost (`model_pricing.py`, `token_usage_utils.py`)
- Uses the SDK's `ModelPricing` type; removes the local `Pricing` schema (`common/schemas/dial.py` deleted).
- Coerces DIAL's string prices to floats with warning-on-failure handling.

### Consumers migrated off raw httpx
- Admin channel download (`routers/channel.py`) and import/export jobs (`import_export_jobs.py`) now stream via the SDK; imports are buffered to a temp file (`aiofiles`) before upload.
- App pricing loader (`channel_completion.py`) and role checks (`auth_context.py`) use `AsyncDial`; role check denies by default when user info can't be fetched/parsed.

### Dependencies (`pyproject.toml`, `poetry.lock`, `.ort.yml`)
- Adds `aidial-client >=0.14.0`; removes `httpcore` as a direct dep; promotes `httpx` to a main dependency.
- Adds ORT license resolution for `aidial-client` (Apache 2.0).

## Checklist


statgpt/admin/routers/channel.py
```diff
@@ -180,18 +179,17 @@ async def download_job_result_by_id(
...
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Job with id={job_id} has no associated file",
+        )
+
+    async with dial_core_factory(
+        base_url=dial_settings.url, api_key=SystemUserAuthContext().api_key
+    ) as dial_core:
+        content, media_type = await dial_core.get_file_with_type(job.file)
+    return Response(content=content, media_type=media_type)
```
Fedir-Yatsenko: ### Export download endpoint no longer streams — regression for large zips

`get_file_with_type` calls `download.aget_content()` (full `aread()`), so the entire exported channel zip is buffered into memory inside the admin pod before the response starts. Export archives can easily be tens-to-hundreds of MB. The PR added `write_file_to` precisely to avoid this in `import_export_jobs.download_zip_file` — please reuse the same pattern here, e.g. stream via `StreamingResponse(...)` or expose a streaming variant on `DialCore`.

statgpt/common/utils/dial/core.py
```diff
@@ -45,213 +48,187 @@ async def read_file_with_progress(file_path: str, chunk_size: int = 64 * 1024) -
...
-            files={name: (name, content, mime_type)},
+            bucket = await self._dial.my_bucket()
+        metadata = await self._dial.files.upload(
+            f"files/{bucket}/{name}", file=(name, content, mime_type)
         )
-        response.raise_for_status()
-        return response.json()
+        result = metadata.model_dump(by_alias=True)
+        result['contentLength'] = metadata.content_length or 0
+        result['contentType'] = metadata.content_type or ''
```
Fedir-Yatsenko: Why are these two lines (118-119) needed?

Fedir-Yatsenko: And on lines 145-146

statgpt/common/utils/dial/core.py
Fedir-Yatsenko, line 1: # Does `DialCore` still earn its keep?

After this refactor, most of `DialCore` is a dict-returning shim around `AsyncDial`:

- `delete_file`, `get_file`, `get_bucket` are pure pass-throughs.
- `put_file`, `put_local_file`, `get_model_by` do `model_dump(by_alias=True)` and then patch the dict — actively throwing away the typed `FileMetadata` / `ModelInfo` the SDK hands us.
- Only three methods do something the SDK can't trivially do today: `get_user_info`, paginated `get_file_metadata` (with `updatedAt` + `nextToken`), and the streaming `write_file_to`. I haven't exhaustively read the SDK; some may already be covered by APIs I missed.

A couple of `aidial-client` capabilities that are worth knowing about, since they affect how we wrap it:

- **Auth can be a callable.** `AsyncAuthValue` accepts `Callable[[], Awaitable[str]]` (`aidial_client/_auth.py:5`), and `AsyncHTTPClient.request` re-invokes it on every request (`_http_client/_async.py:56`). So a single `AsyncDial` can serve many users by passing a `contextvars`-backed callable for `api_key`.
- **A shared connection pool is already supported.** `AsyncDialClientPool` (`aidial_client/_client_pool.py`) owns one `httpx.AsyncClient` and injects it into each `AsyncDial` it creates. And even without the pool, `AsyncHTTPClient` accepts an `internal_http_client` argument directly — so we can hold one `httpx.AsyncClient` at app-lifespan scope, build `AsyncDial` instances per call cheaply, and own teardown ourselves. That removes the need for the `dial._http_client.internal_http_client.aclose()` workaround in `dial_core_factory` entirely.

## Proposal

Happy to keep `DialCore` for this PR if it scopes the change, but I'd like us to commit to retiring it as a follow-up:

1. Manage the `httpx.AsyncClient` ourselves — either:
   - **App-lifespan scope:** hold one shared `httpx.AsyncClient` (or `AsyncDialClientPool`) for the life of the process, so the connection pool is reused across requests.
   - **Per-scope:** create a scoped `httpx.AsyncClient` with `async with` for the duration of a single operation/request and pass it via `AsyncHTTPClient(internal_http_client=...)`. Same shape as today's `dial_core_factory`, but the close happens through httpx's own context manager instead of touching `aidial-client` internals.

   The app-scope variant is the cleaner long-term answer (one connection pool, no per-call setup), but the per-scope variant is a smaller diff and removes the private-attribute `aclose` immediately.
2. Build `AsyncDial` instances per call on top of whichever variant we pick; let callers consume the typed `FileMetadata` / `ModelInfo` directly instead of the dict-massaged returns.
3. For the three remaining bespoke endpoints (user info, paginated metadata, streaming download), first confirm the SDK doesn't already cover them. If it doesn't, prefer raising the gap upstream — only fall back to a small local free function if the SDK can't deliver the fix on a timeline that works for us.

## Worth raising upstream

These are the gaps I think are still real after the points above. Please double-check against the latest SDK before filing — I may still be missing an API:

1. **`FileDownloadResponse` should expose response metadata.** We touch `download._response.headers["content-type"]` in `get_file_with_type` because `FileDownloadResponse` only exposes `aget_content` / `aiter_bytes` / `filename`. A public `content_type` / `headers` accessor would close it.
2. **Truly streaming downloads.** `FileDownloadResponse.aiter_bytes()` exists, but `AsyncHTTPClient.request` (`aidial_client/_http_client/_async.py:60`) does `await self._internal_http_client.send(request)` without `stream=True`, so the body is fully buffered before we iterate. Worth confirming with the maintainers — it may be intentional, or there may be a separate streaming entry point I didn't find.
3. **Lifecycle/teardown on `AsyncDialClientPool` (and arguably `AsyncDial`).** Neither exposes `aclose()` / `__aenter__` / `__aexit__`. Not blocking us — we can manage the underlying `httpx.AsyncClient` ourselves via `internal_http_client` injection — but worth raising as a polish item so consumers don't have to bypass the pool for a clean shutdown.

statgpt/admin/routers/channel.py
kryachkow: fixed, added streaming download

statgpt/common/utils/dial/core.py
kryachkow: core.py is deleted

statgpt/common/utils/dial/core.py
kryachkow: issues resolved on dialclient repo and functionality is used in current pr

statgpt/common/utils/dial/core.py
```diff
@@ -1,257 +0,0 @@
...
-
-    async def get_bucket_json(self) -> dict[str, str]:
-        response = await self._client.get('/v1/bucket')
-        response.raise_for_status()
-        return response.json()
-
-    async def load_bucket(self) -> str:
-        bucket_json = await self.get_bucket_json()
-        if "appdata" in bucket_json:
-            return bucket_json["appdata"]
```
Fedir-Yatsenko, line 100: Could you please bring back this behavior?

statgpt/common/utils/dial/attachments.py
```diff
@@ -1,17 +1,57 @@
...
+                    f"Uploaded {uploaded / (1024 * 1024):.2f} MB / "
+                    f"{file_size / (1024 * 1024):.2f} MB ({percent:.1f}%)"
+                )
+
+    _log.info(f"Upload completed: {uploaded / (1024 * 1024):.2f} MB total")
+    buffer.seek(0)
+    return buffer
 
 
 class AttachmentResponse(BaseModel):
```
Fedir-Yatsenko: Do we still need this class?

statgpt/common/utils/dial/model_pricing.py
```diff
@@ -41,16 +42,16 @@ async def get_model_pricing(self, model: str) -> Pricing | None:
...
             logger.error(f"Failed to fetch model data for model {model}: {e}")
             return None
 
-        if "pricing" not in model_data:
+        if model_data.pricing is None:
             return None
 
         try:
-            return Pricing.model_validate(model_data["pricing"])
+            return Pricing.model_validate(model_data.pricing.model_dump())
```
Fedir-Yatsenko: And this one?

statgpt/common/utils/dial/core.py
kryachkow: functionality returned through resolve_bucket method

statgpt/common/utils/dial/attachments.py
kryachkow: Yes. But it's no longer a duplicate of `FileItem` — it subclasses it and adds the one thing aidial-client doesn't type: `updated_at`.

DIAL returns `updatedAt` only on folder-listing entries, and `FileItem` keeps it as an untyped model *extra* (it's an `ExtraAllowModel`). We filter on it in `_delete_dial_files`:

```python
if file.updated_at is not None and file.updated_at < to_date_timestamp:

statgpt/common/utils/dial/model_pricing.py
kryachkow: Pricing class removed

Fedir-Yatsenko:
/deploy-review

>GitHub actions run: [28654325855](https://github.com/epam/ai-dial-ci/actions/runs/28654325855)

| [deploy-review](https://chat-statgpt-backend-pr-378.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/28654325855/job/84979623962) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/28654325855/job/84980718219) |

statgpt/common/utils/dial/attachments.py
Fedir-Yatsenko: Makes sense to keep it for now. But should we open an issue on `aidial-client` to type this properly upstream?

The SDK already declares `updated_at: int` on `ConversationItem` and `PromptItem` — `FileItem` is the only listing type missing it, even though DIAL returns `updatedAt` for file folder-listings the same way. So this looks like an oversight rather than an intentional gap.

If they add `updated_at: int | None = None` to `FileItem` (`| None` because upload responses don't carry it), we could drop this subclass entirely and rely on the typed field. Right now the whole thing rides on `extra="allow"` — if that ever tightens, `updated_at` silently becomes always-`None` with no type error.

Could you please file the issue and link it here?

kryachkow: https://github.com/epam/ai-dial-client-python/issues/116

kryachkow:
/deploy-review

>GitHub actions run: [29081977410](https://github.com/epam/ai-dial-ci/actions/runs/29081977410)

| [deploy-review](https://chat-statgpt-backend-pr-378.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29081977410/job/86326695697) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/29081977410/job/86328097141) |

Fedir-Yatsenko:
/deploy-review

>GitHub actions run: [29084564111](https://github.com/epam/ai-dial-ci/actions/runs/29084564111)

| [deploy-review](https://chat-statgpt-backend-pr-378.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29084564111/job/86335104516) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/29084564111/job/86336457805) |

kryachkow:
/deploy-review

>GitHub actions run: [29233383141](https://github.com/epam/ai-dial-ci/actions/runs/29233383141)

| [deploy-review](https://chat-statgpt-backend-pr-378.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29233383141/job/86762405282) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/29233383141/job/86764550028) |

Fedir-Yatsenko (approved):

PR #385, opened 2026-05-20
fix: add change detection for channels, data sources, and datasets in content management

### LINKED ISSUE #204: Fix the `content init` command on the CLI
### StatGPT Backend version

latest

### Short description

The `content init` command on the command line always updates the configuration, even if there are no changes.

Run `content init` command in the command line.

Updates should only be made if there are any changes.

### What do you see instead?

Updates are always performed.

### Additional information

There is code that should prevent this, but it doesn't work. This is probably due to missing default values ​​in the configuration, but is returned by the admin app.

kryachkow:
## Applicable issues

#204

### `statgpt/cli/commands/content.py`
Added change-detection to the `content init` command so that channels, data sources, and datasets are only updated when their configuration has actually changed.
Previously every run unconditionally called the update API for all existing entities. Now each entity type has a dedicated comparison helper:
- **`_channel_changed`** — validates the incoming YAML config through `ChannelBase` (filling in all Pydantic defaults via `ChannelConfig`) and compares a canonical `model_dump` against the stored channel.
- **`_data_source_changed`** — normalises the incoming `details` dict through the data-source-type Pydantic config model (looked up via `DataManager.get_config_class`) and compares all updatable fields.
- **`_dataset_changed`** — normalises `details` through the handler's `parse_data_set_config` (looked up via `DataManager.get_data_source_handler_class`). When `use_title_from_src=True` the `title` field is excluded from the comparison to avoid an update loop where the CLI writes the YAML title and the SDMX source immediately restores its own title.
All three helpers fall back to `return True` (always update) if normalisation fails, so the behaviour is safe against unexpected input. The call sites in `_process_channels`, `_process_data_sources`, and `_process_datasets` now print `unchanged` instead of `Updated` when nothing differs.

## Checklist


statgpt/cli/commands/content.py
```diff
@@ -334,6 +338,73 @@ def _get_mime_type(filename: str) -> str:
...
 
+def _channel_changed(incoming_cfg: dict[str, Any], existing: Channel) -> bool:
+    """Return True if channel config differs from what is stored."""
+    try:
+        normalized = ChannelBase.model_validate(incoming_cfg)
+        incoming_dump = normalized.model_dump(mode='json', by_alias=True)
+        existing_dump = existing.model_dump(
+            mode='json',
+            by_alias=True,
+            include={'title', 'description', 'deployment_id', 'llm_model', 'details'},
```
Fedir-Yatsenko: If we convert both configs to `ChannelBase` and compare the Pydantic models, is this a more robust option than hardcoding the exact fields?

Fedir-Yatsenko: The same can be done for data sources, but it won't work for datasets. This will require some adjustments, but even so, we can still avoid hardcoding all fields by using the `exclude` argument for `title`.

statgpt/cli/commands/content.py
```diff
@@ -334,6 +338,73 @@ def _get_mime_type(filename: str) -> str:
...
+) -> bool:
+    """Return True if dataset config differs from what is stored."""
+    try:
+        incoming_details = incoming_cfg.get('details', existing.details)
+        if data_source is not None:
+            handler_class = DataManager.get_data_source_handler_class(data_source.type.name)
+            parsed_config = handler_class.parse_data_set_config(incoming_details)
+            normalized_details = parsed_config.model_dump(mode='json', by_alias=True)
+            # When the data source manages the title automatically, skip title comparison
+            # to avoid an update loop (CLI sets YAML title → source restores its own title).
```
Fedir-Yatsenko, line 396: Good catch👍 This is likely the cause of the unexpected behavior. It looks like we need to rethink our approach to the `title` field, but that's outside the scope of this PR/issue.

statgpt/cli/commands/content.py
kryachkow: refactored

statgpt/cli/commands/content.py
```diff
@@ -341,14 +341,9 @@ def _get_mime_type(filename: str) -> str:
...
-        normalized = ChannelBase.model_validate(incoming_cfg)
-        incoming_dump = normalized.model_dump(mode='json', by_alias=True)
-        existing_dump = existing.model_dump(
-            mode='json',
-            by_alias=True,
-            include={'title', 'description', 'deployment_id', 'llm_model', 'details'},
-        )
-        return incoming_dump != existing_dump
+        incoming = ChannelBase.model_validate(incoming_cfg)
+        stored = ChannelBase.model_validate(existing.model_dump(mode='json', by_alias=True))
```
Fedir-Yatsenko, line 353: `model_dump` is not needed, we can do the following: `ChannelBase.model_validate(existing, from_attributes=False)`

statgpt/cli/commands/content.py
```diff
@@ -334,6 +338,62 @@ def _get_mime_type(filename: str) -> str:
...
+
+def _data_source_changed(incoming_cfg: dict[str, Any], existing: DataSource) -> bool:
+    """Return True if data source config differs from what is stored."""
+    try:
+        config_class = DataManager.get_config_class(existing.type.name)
+        incoming_details = incoming_cfg.get('details', existing.details)
+        normalized_details = config_class(**incoming_details).model_dump(mode='json', by_alias=True)
+        incoming = DataSourceBase.model_validate({**incoming_cfg, 'details': normalized_details})
+        stored = DataSourceBase.model_validate(existing.model_dump(mode='json', by_alias=True))
+        return incoming != stored
```
Fedir-Yatsenko, line 375: Shouldn't this method check not only the "details" field, but also the "description"?

kryachkow: Yes, it should, and with the current implementation it does.

_data_source_changed() builds a DataSourceBase for both sides:

incoming = _as_schema(DataSourceBase, {**incoming_cfg, 'details': normalized_details})
stored = _as_schema(DataSourceBase, existing)
return incoming != stored
Since DataSourceBase includes title, description, type_id, and details, the model comparison includes description automatically. So a changed data source description should now trigger an update without hardcoding the field.

statgpt/cli/commands/content.py
kryachkow: with Pydantic v2, validating a subclass instance returns the original Channel, so equality will fail against ChannelBase

we can put smth like 
`stored = ChannelBase.model_validate(
            {field: getattr(existing, field) for field in ChannelBase.model_fields}
        )
`

but i dont think this is better than current implementation

Fedir-Yatsenko: Wow, I didn't know that, let's leave your version then.

statgpt/cli/commands/content.py
```diff
@@ -334,6 +338,62 @@ def _get_mime_type(filename: str) -> str:
...
+    except Exception:
+        return True
+
+
+def _data_source_changed(incoming_cfg: dict[str, Any], existing: DataSource) -> bool:
+    """Return True if data source config differs from what is stored."""
+    try:
+        config_class = DataManager.get_config_class(existing.type.name)
+        incoming_details = incoming_cfg.get('details', existing.details)
```
Fedir-Yatsenko: If the configuration is missing,  let's stop and display an error message to the user.

statgpt/cli/commands/content.py
```diff
@@ -334,6 +338,62 @@ def _get_mime_type(filename: str) -> str:
...
+        normalized_details = config_class(**incoming_details).model_dump(mode='json', by_alias=True)
+        incoming = DataSourceBase.model_validate({**incoming_cfg, 'details': normalized_details})
+        stored = DataSourceBase.model_validate(existing.model_dump(mode='json', by_alias=True))
+        return incoming != stored
+    except Exception:
+        return True
```
Fedir-Yatsenko, line 379: ### `except Exception: return True` is too broad and silent — and inconsistent with the rest of the CLI

```python
except Exception:
    return True
```
appears three times (`content.py:347, 360, 393`). This violates `CODE_STYLE.md` §10 ("Avoid silent failures — log or propagate exceptions with helpful messages") and is inconsistent with peers in this very file (`_process_client` re-raises with `print_error`, `_upload_dial_files` reports with `print_error`). Today the symptom of a real bug (wrong handler name, broken Pydantic schema, KeyError on `existing.type`) is silently "everything is always changed", which is the opposite of what this PR is trying to deliver.

Concrete suggestion: narrow + log.
```python
except (ValidationError, KeyError, AttributeError) as exc:
    print_warning(f"  change-detection skipped (assuming changed): {exc}")
    return True
```
Anything not in that set should propagate — a programming error in normalisation should not look like a "real" change in the user-facing summary.

statgpt/cli/commands/content.py
```diff
@@ -334,6 +338,62 @@ def _get_mime_type(filename: str) -> str:
...
+
+
+def _data_source_changed(incoming_cfg: dict[str, Any], existing: DataSource) -> bool:
+    """Return True if data source config differs from what is stored."""
+    try:
+        config_class = DataManager.get_config_class(existing.type.name)
+        incoming_details = incoming_cfg.get('details', existing.details)
+        normalized_details = config_class(**incoming_details).model_dump(mode='json', by_alias=True)
```
Fedir-Yatsenko: Line 356 uses the constructor; the other two helpers use `model_validate`. Constructors bypass alias resolution and validators that `model_validate` runs. Prefer:
```python
normalized_details = config_class.model_validate(incoming_details).model_dump(mode='json', by_alias=True)
```
for consistency.

statgpt/cli/commands/content.py
kryachkow: added

statgpt/cli/commands/content.py
kryachkow: fixed

statgpt/cli/commands/content.py
kryachkow: fixed

Fedir-Yatsenko (approved):

PR #399, opened 2026-05-25
fix: incorrect detection of reindexing status in the update endpoint

### LINKED ISSUE #212: [admin] content init falsely reports datasets as needing re-indexing after successful reindex
### StatGPT Backend version

0.5.0

### Short description

`content init` reports `NEEDS_REINDEX` for datasets that have already been successfully re-indexed.

1. Run `content init` — some datasets are reported as needing re-indexing
2. Run `channel reindex` for the affected datasets and wait for indexing to complete
3. Run `content init` again
4. The same datasets are still listed under "Datasets requiring reindexing"

After a successful re-index, `content init` should recognize that the datasets are up to date and not report them as needing re-indexing.

### What do you see instead?

`content init` continues to report the datasets as `NEEDS_REINDEX` regardless of whether they were recently re-indexed.

### Additional information

The re-indexing status is determined in `AdminPortalDataSetService._propagate_config_to_channel_datasets` (`statgpt/admin/services/dataset.py`, lines 722-795). The logic compares the `indexing_hash` computed from the current resolved config against the `indexing_config_hash` stored on the last completed `ChannelDatasetVersion`:

```python
new_config_hash = resolved_parsed.indexing_hash

if new_config_hash == last_completed.indexing_config_hash:
    # AUTO_UPDATED
else:
    # NEEDS_REINDEX
```

The likely issue is that the `indexing_config_hash` stored on the completed version does not match the hash computed from the current config, even though the actual indexed content is equivalent. This could happen if:

- The config resolution produces a slightly different output after re-indexing (e.g., version bumps in URN references, resolved config differences)
- The hash stored during re-indexing was computed from a different config state than what `content init` resolves
- Fields included in `indexing_hash` (via `IndexingField()` markers in `statgpt/common/data/base/indexing.py`) have changed in the config YAML without affecting the actual indexed data

**Relevant code:**
- CLI display logic: `statgpt/cli/commands/content.py` (lines 535-588)
- Re-index decision: `statgpt/admin/services/dataset.py` — `_propagate_config_to_channel_datasets` (lines 722-795)
- Indexing hash computation: `statgpt/common/data/base/indexing.py` — `compute_indexing_hash()`
- Version model: `statgpt/common/models/models.py` — `ChannelDatasetVersion.indexing_config_hash`

Fedir-Yatsenko:
### Applicable issues

- fixes #212 

### Checklist

### DIFF
--- a/statgpt/common/data/sdmx/common/config.py
+++ b/statgpt/common/data/sdmx/common/config.py
@@ -1,6 +1,6 @@
 import logging
 from enum import StrEnum
-from typing import Annotated, Any, Literal
+from typing import Any, Literal
 
 from pydantic import Field
 
@@ -11,7 +11,6 @@ from statgpt.common.data.base import (
     DataSetConfigTemplate,
     DataSetHierarchyConfig,
     DataSourceConfig,
-    IndexingField,
 )
 from statgpt.common.data.sdmx.common.data_explorer_url import DataExplorerUrlConfig
 
@@ -140,9 +139,9 @@ class CategorySchemaDataSetHierarchyConfig(BaseModel):
 class UrnReference(BaseModel):
     """This class represents a URN reference for SDMX datasets and allows dynamic values."""
 
-    agency_id: Annotated[str, IndexingField()]
-    resource_id: Annotated[str, IndexingField()]
-    version: Annotated[str, IndexingField()] = Field(default='latest')
+    agency_id: str
+    resource_id: str
+    version: str = Field(default='latest')
 
     def short_urn(self) -> str:
         """Return a short URN representation."""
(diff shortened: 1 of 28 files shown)


statgpt/common/data/sdmx/common/config.py
```diff
@@ -229,7 +228,7 @@ class SdmxDataSetConfigMixin:
     use_title_from_src: bool = Field(
         default=False, description="Whether to use the title obtained from the source"
     )
-    urn: Annotated[UrnReference, IndexingField()] = Field(description="The URN of the dataset")
```
Fedir-Yatsenko, line 232: TODO: check what if `urn` is changed

Fedir-Yatsenko: Done ✔️

ypldan (approved):

PR #413, opened 2026-05-27
fix: allow detaching dataset when embedding model is unavailable #412

### LINKED ISSUE #412: Detaching a dataset from a channel fails when the embedding model is unavailable
### StatGPT Backend version

latest

### Short description

Removing a dataset from a channel fails when the channel's embedding model deployment is unavailable, even though the cleanup step does not actually need embeddings.

1. Configure a channel with an embedding model deployment that is no longer reachable (e.g., deployment removed, credentials revoked, or DIAL Core returns an error for the configured `llm_model`).
2. Call `DELETE /channels/{channel_id}/datasets/{dataset_id}` for a dataset attached to that channel.
3. Observe the request fails before the `channel_dataset` row is deleted.

The dataset should be detachable from the channel even when the embedding model deployment is unreachable. Vector store documents should still be cleared (deletion does not require embeddings), and the `channel_dataset` row should be removed.

### What do you see instead?

`DataSetService._clear_vector_stores` (`statgpt/admin/services/dataset.py:939`) calls `VectorStoreFactory.get_vector_store(..., embedding_model_name=channel.llm_model)`. `VectorStoreFactory.get_vector_store` (`statgpt/common/vectorstore/factory.py:21`) eagerly resolves the embedding model via `self._embedding_models.get(...)` and passes it into `PgVectorStore.__init__`, even though the subsequent `remove_documents_by` call only runs SQL DELETEs keyed by `dataset_id` / `version_ids` and never uses the embedding model.

As a result, an unreachable embedding deployment makes the factory raise during construction, which aborts vector store cleanup and prevents the `channel_dataset` row from being deleted.

### Additional information

Proposed direction: make the embedding model optional for delete-only usage of the vector store — for example, allow `embedding_model_name=None` in `VectorStoreFactory.get_vector_store` (or add a dedicated delete-only entry point) and make `PgVectorStore.embedding_model` optional, used only by ingest/search paths. `_clear_vector_stores` would then build the store without resolving the embedding model and call `remove_documents_by` as today.

Out of scope: failures in other cleanup steps (e.g., Elastic deletes, DB delete) should still block channel-dataset removal — this issue is strictly about removing the unnecessary embedding-model dependency from the vector store delete path.

Fedir-Yatsenko:
### Applicable issues

- fixes #412

### Description of changes

`DELETE /channels/{channel_id}/datasets/{dataset_id}` failed when the channel's embedding deployment was unreachable: `VectorStoreFactory.get_vector_store` eagerly resolved the embedding model via DIAL Core and raised before any cleanup ran. The actual delete SQL in `PgVectorStore.remove_documents_by` never uses embeddings — only `_get_document_model`'s `Vector(embedding_length)` column tied the path to the deployment.

Split the interface so embedding-free operations don't require DIAL:

- New `VectorStoreEmbeddingFree` protocol exposes only ops that don't need embeddings (`clear`, `remove_documents_by`, `has_duplicates`, size getters, `deduplicate_by_document_content`). `VectorStore` extends it with the embedding-requiring methods.
- New `VectorStoreFactory.get_embedding_free_vector_store(...)` constructs the narrower store without contacting DIAL.
- Migrated admin callers that only use embedding-free ops (`_clear_vector_stores`, `check_index_status`, `_clear_vector_store`, `deduplicate_channel_dimensions`) and dropped the now-unused `auth_context` plumbing along their call chains.

### Checklist

### DIFF
--- a/statgpt/admin/auto_update.py
+++ b/statgpt/admin/auto_update.py
@@ -85,7 +85,7 @@ async def _log_results(job_ids: list[int]) -> bool:
     return failed == 0
 
 
-async def _deduplicate_channels(channel_ids: set[int], auth_context: AuthContext) -> None:
+async def _deduplicate_channels(channel_ids: set[int]) -> None:
     """Run deduplication for channels that had a reindex.
 
     NOTE: The number of concurrent executions is limited by the semaphore
@@ -96,9 +96,7 @@ async def _deduplicate_channels(channel_ids: set[int], auth_context: AuthContext
     _log.info(f"Running deduplication for {len(sorted_ids)} channel(s) with reindex: {sorted_ids}")
     results = await asyncio.gather(
         *(
-            deduplicate_dimensions_in_background_task(
-                channel_id=channel_id, auth_context=auth_context
-            )
+            deduplicate_dimensions_in_background_task(channel_id=channel_id)
             for channel_id in sorted_ids
         ),
         return_exceptions=True,
@@ -128,7 +126,7 @@ async def run_auto_update() -> bool:
 
     reindex_channel_ids = await _get_reindex_channel_ids(job_ids)
     if reindex_channel_ids:
-        await _deduplicate_channels(reindex_channel_ids, auth_context)
+        await _deduplicate_channels(reindex_channel_ids)
 
     return await _log_results(job_ids)
 
--- a/statgpt/admin/routers/channel.py
+++ b/statgpt/admin/routers/channel.py
@@ -238,7 +238,7 @@ async def delete_channel(
 ) -> None:
     """Delete channel by id"""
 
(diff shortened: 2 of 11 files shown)


Fedir-Yatsenko:
/deploy-review

>GitHub actions run: [26510269945](https://github.com/epam/ai-dial-ci/actions/runs/26510269945)

| [deploy-review](https://chat-statgpt-backend-pr-413.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/26510269945/job/78072763121) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/26510269945/job/78081307499) |

ypldan (approved):

PR #417, opened 2026-06-02
chore: temporarily disable ORT checks

Fedir-Yatsenko:
### Applicable issues

N/A

### Description of changes

ORT checks are temporarily disabled until the Ci/CD pipeline is updated.

### Checklist

- [ ] ~Deployed and tested in a `Review` environment.~ N/A

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ on:
       ort-enabled:
         type: boolean
         description: Enable ORT scanning
-        default: true
+        default: false
       ort-bypassed:
         type: boolean
         description: Do not fail pipeline if ORT scan failed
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -26,7 +26,7 @@ on:
       ort-enabled:
         type: boolean
         description: Enable ORT scanning
-        default: true
+        default: false
```
MykhailoRyzhman, line 29: This default value for input is only used if we don't pass anything here through other workflows. However, in the PR and Review workflows, we pass `ort-enabled`.

Let's leave everything as it was before in `test.yml`. Although this isn't critical.

MykhailoRyzhman (approved):

.github/workflows/test.yml
Fedir-Yatsenko: Changed just in case any pipeline uses the default value.

PR #508, opened 2026-07-02
feat: run out-of-scope guardrail concurrently with the supreme agent #497

### LINKED ISSUE #497: Run the out-of-scope guardrail concurrently with the supreme agent
### StatGPT Backend version

0.12.0 (development)

Every chat request pays a blocking out-of-scope guardrail LLM call before the supreme agent is allowed to start. This adds a full LLM round-trip to time-to-first-token on 100% of requests, while the verdict is almost always "in scope".

Run the guardrail check and the agent concurrently. The agent executes speculatively: user-visible output is buffered, history/state mutations happen on copies, and real tool dispatch is gated on the verdict. On "in scope" the buffer is flushed and the run is committed; on "out of scope" the agent task is cancelled and discarded, and the guardrail response streams as today. Guarded by an env kill switch.

A smaller/faster guardrail model (still serial), or skipping the check for established conversations (changes guardrail semantics).

ypldan:
### Applicable issues

- fixes #497

### Description of changes

The out-of-scope checker ran before the supreme agent, adding an LLM round-trip to time-to-first-token even though verdicts are almost always "in scope". When a channel opts in, the checker and agent now run concurrently: the agent runs speculatively against a recording choice, a copied history, and a deepcopied state, and only commits once the verdict confirms "in scope". Out of scope, the speculative run is cancelled and its output discarded, so the user sees exactly the sequential flow's output — minus one round-trip of latency.

- **Speculation** — `RecordingChoice`/`RecordingStage` record every choice write; on an in-scope verdict the buffer is flushed to the real choice and history/state committed, otherwise it is discarded. Irreversible side effects (real tool dispatch) wait on a permission gate.
- **Isolation by substitution** — the agent only ever holds speculative surfaces (recording choice, copied history/state, `adopt_stage`-wrapped performance stage), so nothing can leak to the real choice before commit. The substitution is built in one place (`_build_speculative_inputs`) and a unit test guards the invariant by rejecting any un-substituted choice/stage surface on the speculative inputs.
- **Cleanup guarantee** — an `async with` context manager reaps both tasks and drops any un-flushed output on every exit (commit, out-of-scope, checker failure, cancellation, or a mid-replay flush failure).
- **Observability** — committed vs discarded speculation is logged; a speculative agent failure on the out-of-scope path is surfaced rather than swallowed.
- **Per-channel toggle** — `out_of_scope.optimistic` (default off); off keeps the exact sequential check-then-agent chain.

### Checklist


ypldan:
/deploy-review

>GitHub actions run: [29020787991](https://github.com/epam/ai-dial-ci/actions/runs/29020787991)

| [deploy-review](https://chat-statgpt-backend-pr-508.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29020787991/job/86126800376) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/29020787991/job/86131526841) |

PR #528, opened 2026-07-14
feat: merge an import into an existing channel without cleaning it up #527

### LINKED ISSUE #527: Support merging an import into an existing channel without cleaning it up
### StatGPT Backend version

latest

Refreshing a live channel from an updated archive currently forces a **full clean-up**: the only reliable way to re-import into an existing channel is `clean_up=True`, which drops the whole channel — config, datasets, indexes, and vector stores — and rebuilds it from scratch. That means downtime and an expensive full re-index every time upstream data changes.

Trying to import **without** cleaning up doesn't work as a merge either:

- **Channel config** is always re-created (`ChannelService.import_channel_from_zip` always calls `create_channel`), so a non-clean-up import fails on the duplicate `deployment_id`. There is no in-place update path for an existing channel config.
- **Data source config** for an existing data source is only updated when `update_data_sources=True`, disconnected from a single coherent "merge" operation.
- **Indexes** are added additively on top of existing indexes (`_import_vector_store_tables` / `_import_elastic_data`), but **no deduplication runs afterward**, so appended indicators/dimensions accumulate duplicates and degrade search quality.

We want to import into an existing channel as a **merge that reconciles the archive against the current channel — without dropping and rebuilding it.**

Make import of an existing channel (matched by `deployment_id`) a **merge that never cleans up the channel**:

1. **Channel & data source configs** — update the existing channel config and data source configs in place instead of re-creating them.
2. **Existing datasets** — update in place (extends the current `update_datasets` merge in `DataSetService._import_datasets`).
3. **New datasets** — add datasets present in the archive but not in the channel (already supported; keep).
4. **Indexes** — keep adding archive indexes on top of existing indexes (already additive), no clean-up of existing indexes.
5. **Deduplication after merge** — when the channel already had indexes and the import adds more, run deduplication (`deduplicate_channel_dimensions` for vector stores, and the equivalent for the Elastic indicator/matching indexes) so appended documents don't create duplicates.
6. **Audit logging** — every update/create performed during the merge (channel config, data source config, datasets) must be recorded via the existing `@audit_action` audit log, so the merge leaves a full trail of what changed.

kryachkow:
### Applicable issues

Resolves #527

### Description of changes

Re-importing an archive into a live channel previously required `clean_up=True`, which drops the whole channel (config, datasets, indexes, vector stores) and rebuilds it from scratch — causing downtime and a full re-index. Importing without cleanup wasn't a real merge either: `create_channel` was always called, so a non-clean-up import failed on the duplicate `deployment_id`.

This PR makes a non-clean-up import (matched by `deployment_id`) a **merge** that reconciles the archive against the current channel instead of dropping it:

- **Channel config** — updated in place via the audited `update()` when the channel already exists, instead of re-creating it. `clean_up=True` keeps the drop-and-rebuild behavior.
- **Datasets** — on merge, existing datasets are always reconciled in place; new datasets from the archive are added. Fixed `_add_datasets_to_channel` to skip already-linked datasets (`ChannelDataset` has no unique constraint, so a merge would otherwise create duplicate links/versions).
- **Data source config** — stays gated behind the existing `update_data_sources` flag.
- **Indexes** — the non-indicator and special dimension vector stores are appended additively on merge (new `clear_existing` flag on `import_from_zipfile`); the indicator collection is still rebuilt from the full-channel archive since it isn't covered by the dedup pass.
- **Deduplication after merge** — after a merge that imports indexes, `deduplicate_channel_dimensions` runs so appended documents don't accumulate duplicates. A dedup failure is logged but does not fail the import.
- **Audit logging** — every create/update during the merge goes through the existing `@audit_action`-decorated methods, so the merge leaves a full change trail.

Elastic indicator/matching indexes need no extra dedup step: `add_bulk` upserts by document `id`, so re-importing the same document overwrites rather than duplicates.

Verified with `mypy`, `flake8`, `black`, and the admin unit tests. Note: no end-to-end integration test for the merge flow was added (it requires DB + vector store + Elastic containers).

### Checklist

### DIFF
--- a/statgpt/admin/routers/channel.py
+++ b/statgpt/admin/routers/channel.py
@@ -89,8 +89,11 @@ async def export_channel(
 
 
 IMPORT_CHANNEL_CLEAN_UP_DESCRIPTION = (
-    "If enabled and a channel with the same `deployment_id` exists, it will be deleted."
-    " If disabled and a channel with the same `deployment_id` exists, the import job will fail."
+    "If enabled and a channel with the same `deployment_id` exists, it will be deleted"
+    " and rebuilt from scratch."
+    " If disabled and a channel with the same `deployment_id` exists, the archive is merged"
+    " into it: configs and datasets are updated in place, new datasets and indexes are added,"
+    " and dimension indexes are deduplicated afterwards."
 )
 
 
--- a/statgpt/admin/services/channel.py
+++ b/statgpt/admin/services/channel.py
@@ -205,10 +205,9 @@ class AdminPortalChannelService(ChannelService):
         zip_file: zipfile.ZipFile,
         clean_up: bool,
         scope: schemas.ExportScope,
-        deployment_id: str | None,
+        deployment_id: str,
         auth_context: AuthContext,
-        existing_channel: models.Channel | None = None,
-    ) -> models.Channel:
+    ) -> tuple[bool, models.Channel]:
         if scope.includes_dial_files():
             await self._import_dial_files_from_zip(zip_file, auth_context)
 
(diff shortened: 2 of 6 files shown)


Fedir-Yatsenko:
/deploy-review

>GitHub actions run: [29494987795](https://github.com/epam/ai-dial-ci/actions/runs/29494987795)

| [deploy-review](https://chat-statgpt-backend-pr-528.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29494987795/job/87609716573) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/29494987795/job/87611219701) |

Fedir-Yatsenko (approved):

PR #563, opened 2026-07-29
fix: handle invalid message history gracefully

Fedir-Yatsenko:
## Applicable issues

None

## Description of changes

A request with an empty user message used to fail deep inside the out-of-scope checker with a bare `ValueError`, which the generic `except Exception` turned into a traceback plus a `200 OK` "An error occurred while processing your request."

`History` is now built at the request boundary (in `ChannelCompletion`, not via a `RunnablePassthrough` step in the main chain) and validates user messages up front, raising a new domain exception `InvalidHistoryError`. `ChannelCompletion` catches it, logs a single warning without a traceback, and answers "This conversation contains a message that cannot be processed. Please start a new chat." — starting a new chat, because the offending message stays in the history the client resends.

For debugging, that handler also dumps the request messages into `state['received_messages']`, trimmed of assistant `custom_content` and attachment payloads. This is why the reply is a `200`: the SDK discards the accumulated response as soon as a `DIALException` reaches the stream, which would throw the dump away.

Also added `except DIALException: raise`, so the `InvalidRequestError` (400) that `SystemMessageInterceptor` already raises is no longer downgraded to the generic 200 message.

## Checklist

### DIFF
--- a/statgpt/app/application/channel_completion.py
+++ b/statgpt/app/application/channel_completion.py
@@ -19,6 +19,7 @@ from statgpt.app.services.chat_facade import ChannelServiceFacade
 from statgpt.app.settings.dial_app import dial_app_settings
 from statgpt.app.utils.dial_exceptions import RateLimitException
 from statgpt.app.utils.dial_stages import optional_timed_stage
+from statgpt.app.utils.message_history import History, InvalidHistoryError, dump_dial_messages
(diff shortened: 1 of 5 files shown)


Fedir-Yatsenko:
<details>

<summary>Example response</summary>

image

</details>

kryachkow (dismissed):

kryachkow (approved):

Issue #490, opened 2026-07-01
Add Deep Research tool


Fedir-Yatsenko:
### What is the problem this feature will solve?

The agent can answer targeted questions, but it cannot perform open-ended, multi-source investigation over statistical data and publications. Users need a way to ask broad research questions and get a synthesized, grounded answer that draws on multiple datasets and sources.

### What is the proposed feature or solution?

Add a **Deep Research** agent tool that decomposes a research question, gathers evidence across existing data/publication sources, and synthesizes a grounded, cited report.

Link: https://github.com/epam/ai-dial-deep-research-backend

TODO:
- [x] Create a basis for a new tool and configuration.
- [x] Provide streaming of `stages` and `attachments`.
- [x] Add saving of conversation between the Supreme Agent and the Deep Research to `state`.
- [x] New user messages should be forwarded to the Deep research, until it is completed. (Clarification flow) We need to discuss a specific mechanism.
- [x] Deep Research mode shall be activated exclusively by user clicking on a dedicated UI control (This needs to be implemented on the frontend side first.). This button forces the Supreme Agent to call the Deep Research tool (achievable via `tool_choice` parameter, see [DIAL API](https://dialx.ai/dial_api#operation/sendChatCompletionRequest)).

### What alternatives have you considered?

N/A

PR #565, opened 2026-07-31
feat: mediate Deep Research clarification flow via Supreme Agent

### LINKED ISSUE #544: Deep Research: mediate clarification flow through the Supreme Agent
### StatGPT Backend version

latest

Sub-issue of #490 (Add Deep Research tool).

The first version of the tool (see #490, "basis" sub-issue) is single-shot: the Supreme Agent calls Deep Research with one `query` argument and streams the report back. But Deep Research is **interactive** — before running it builds a plan, asks the user clarifying questions, and asks the user to approve the plan. Supporting this means carrying a multi-turn conversation between the user and Deep Research across several user messages, until Deep Research completes.

**Short answer / core requirement:** this communication must be **mediated by the Supreme Agent**, not a blind forward of messages between Deep Research and the user. The Supreme Agent must stay the single owner of the chat and its `state`, decide whether an incoming user message continues the Deep Research session or starts a new request, and be able to answer some clarifications itself from conversation context without bothering the user.

This sub-issue covers these TODOs of #490:
- Save the conversation between the Supreme Agent and Deep Research to `state`.
- Forward subsequent user messages to Deep Research until it completes (clarification flow) — mechanism to be agreed.

**1. Persist the conversation in `state`**
- Store the Supreme Agent ↔ Deep Research exchange (questions returned, user answers, plan, approval) in DIAL `state`. The codebase already has an entry point for storing this "conversation between" — nothing blocks us from starting here.

**2. Deep Research session state / flag**
- Track a session status: `in progress` vs `completed`. This flag drives how the next user message is routed.

**3. Message routing (mediated, not forwarded)**
- While a session is `in progress`, the next user message is treated as a continuation of the Deep Research clarification (answer to a question / plan approval), not as a fresh Supreme Agent request.
- The Supreme Agent may answer a clarifying question itself using conversation context, and only surface a question to the user when it genuinely needs input.
- When the session is `completed`, subsequent messages go to the Supreme Agent as usual.

- **Disabled control:** when Deep Research UI control is disabled the tool is fully unavailable to Supreme Agent
- **Multiple runs:** no more than one Deep Research tool run can happen in a single chat parallelly the error should be shown (to discuss: should we allow to use deep research second time after first iteration, if yes how context should be managed)
### To clarify (mechanism / open questions)

- **Final report format:** is it an attachment (Canvas)? If it is not an attachment, how do we deliver it to the user without re-spending tokens or introducing hallucinations? (The final message rendering depends on this.)

### Edge cases to discuss

- User asks to call a **different tool** while a Deep Research session is active → the LLM should refuse and keep the session focused.
- User **ignores / abandons / cancels** a clarifying question mid-session → how to exit the session and reset the flag.
- Deep Research **over-asks** clarifications (observed: it re-asks even when the time range was already provided) — how much should the Supreme Agent absorb vs pass through.
- Ambiguity of the **next user message**: is it an answer to Deep Research, or an unrelated new request for the Supreme Agent? The routing flag must resolve this deterministically.
- The dedicated **"force Deep Research" button** forces the tool via `tool_choice`; the Supreme Agent still has to produce a correct `query` argument.

- **Blind message forwarding** between the user and Deep Research (pass Deep Research output straight to the user and the user's reply straight back). Rejected: the Supreme Agent must own the conversation and `state`, mediate clarifications, and control when the session ends.

kryachkow:
## Applicable issues

- closes #544

## Description of changes

The first version of the Deep Research tool (#490) was single-shot. But Deep Research is **interactive**: before running it clarifies the request and agrees a plan with the user. This PR makes the Supreme Agent **mediate** that multi-turn flow instead of blindly forwarding messages — the agent stays the single owner of the chat and its `state`.

Key points:

- **Session persisted in `state`.** A `DeepResearchSession` (status + the user/assistant sub-conversation, each assistant turn carrying Deep Research's own `custom_content.state`) is stored in DIAL `state` and carried across turns, so Deep Research resumes its preparation flow. A completed session is dropped (the report already lives in chat history); a new query starts fresh.
- **Mediated routing.** While a session is in progress the next user message is routed back into Deep Research as a continuation, not a new request. A per-run (non-persisted) system instruction tells the agent it is relaying between the user and Deep Research.
- **Auto-answering clarifications.** The conversation is summarized once and forwarded to Deep Research as context. Deep Research's clarifying questions are triaged by an LLM: questions answerable from context are answered silently on the user's behalf (looping up to `MAX_CLARIFICATION_ROUNDS = 4`); only the remainder are surfaced to the user, verbatim. Held-back auto-answers are merged with the user's reply on the next turn so Deep Research gets every answer at once.
- **`deep_research` config toggle.** When on, the tool is forced via `tool_choice`; when off, it is hidden from the agent entirely. Turning it off mid-session abandons the session and handles the message normally.
- **Streaming.** Deep Research output is buffered so clarifications can be filtered/consumed, then switches to live streaming the moment `research_started` is signalled, so the final report renders as-is.
- **Single run per chat.** A guard rejects duplicate Deep Research calls within one agent response.

## Implementation details to discuss 

<details>

<summary>Open points worth a review</summary>

- **`research_started` contract.** We treat the session as finished only when Deep Research's persisted `preparation.research_started` is `True`. This assumes the deployment sets that flag **only after** the final report has fully streamed in the same response. If a deployment emits it earlier, the session would close prematurely and the report never resume. Is this contract safe to rely on?
- **State shape coupling.** We read Deep Research's internal `preparation.clarification.questions` / `research_started` out of its opaque `custom_content.state`. This couples us to the deployment's internal schema — worth agreeing how stable that is / whether we need a version guard.
- **Round cap (`MAX_CLARIFICATION_ROUNDS = 4`).** On the final round any still-answerable questions are handed to the user instead of auto-answered, to avoid an endless answer/re-ask loop. Is 4 the right ceiling?
- **Triage aggressiveness.** The triage prompt errs toward asking the user when in doubt. Tuning this trades off "don't bother the user" vs "don't guess wrong" — needs real-conversation validation.
- **Error handling.** Both the tool (mid-stream/request errors) and the agent (framework-level errors) surface a shared `DEEP_RESEARCH_ERROR_MESSAGE` and leave the session untouched. Under forced `tool_choice`, the agent loop ends on a Deep Research result rather than re-invoking, to avoid spinning to `max_agent_iterations`.
- **Config sync.** `StatGPTConfiguration.deep_research` must stay in sync with the `DeepResearchChannelConfiguration` that advertises the toggle to the frontend only when the tool is enabled.

</details>

## Checklist

### DIFF
--- a/statgpt/app/application/channel_completion.py
+++ b/statgpt/app/application/channel_completion.py
@@ -213,7 +213,11 @@ class ChannelCompletion(ChatCompletion):
         if custom_content := last_response.custom_content:
             if state := custom_content.state:
                 if (show_debug_stages := state.get(StateVarsConfig.SHOW_DEBUG_STAGES)) is not None:
-                    return {**defaults, StateVarsConfig.SHOW_DEBUG_STAGES: show_debug_stages}
+                    defaults[StateVarsConfig.SHOW_DEBUG_STAGES] = show_debug_stages
+                # Carry the Deep Research clarification session across turns so the Supreme
+                # Agent can mediate the multi-turn flow and resume the Deep Research run.
+                if (dr_session := state.get(StateVarsConfig.DEEP_RESEARCH_SESSION)) is not None:
+                    defaults[StateVarsConfig.DEEP_RESEARCH_SESSION] = dr_session
 
         return defaults
 
--- a/statgpt/app/chains/deep_research/deep_research_tool.py
+++ b/statgpt/app/chains/deep_research/deep_research_tool.py
@@ -1,29 +1,54 @@
 import time
-from typing import Annotated, Any
+from typing import Annotated, Any, ClassVar
 
 from mcp.types import ToolAnnotations
-from openai import APIError
 from pydantic import Field
 
+from statgpt.app.chains.deep_research.mediation import (
+    build_transcript,
+    format_auto_answers,
+    summarize_conversation,
+    triage_questions,
+)
 from statgpt.app.chains.parameters import ChainParameters
 from statgpt.app.chains.tools import GuardrailInput, StatGptTool, ToolArgs
 from statgpt.app.config import StateVarsConfig
-from statgpt.app.schemas import ToolArtifact, ToolMessageState
+from statgpt.app.schemas import (
+    DEEP_RESEARCH_ERROR_MESSAGE,
+    DeepResearchSession,
+    DeepResearchStatus,
+    ToolArtifact,
+    ToolMessageState,
+)
 from statgpt.app.utils import OpenAiToDialStreamer, openai
+from statgpt.app.utils.dial_stages import ChoiceI
+from statgpt.common.auth.auth_context import AuthContext
 from statgpt.common.config import multiline_logger as logger
 from statgpt.common.schemas import DeepResearchTool as DeepResearchToolConfig
-from statgpt.common.schemas import ToolTypes
+from statgpt.common.schemas import LLMModelConfig, ToolTypes
 from statgpt.common.schemas.llm_call_duration import LLMCallDurationItem
+from statgpt.common.schemas.tool_details import DeepResearchDetails
 from statgpt.common.utils.llm_call_duration_context import get_llm_call_duration_manager
+from statgpt.common.utils.markdown import format_as_markdown_list
 
 
 class DeepResearchArgs(ToolArgs):
     query: Annotated[str, GuardrailInput] = Field(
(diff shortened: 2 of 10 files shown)


Fedir-Yatsenko:
This pull request has been split into two separate PRs:
- https://github.com/epam/statgpt-backend/pull/554
- https://github.com/epam/statgpt-backend/pull/576

**Since they have been merged, I am closing this PR  as a duplicate.**

PR #570, opened 2026-08-03
feat: manage SDMX proxy configuration from the Admin Portal

### LINKED ISSUE #569: Manage SDMX proxy configuration from the Admin Portal
## StatGPT Backend version

latest

## What is the problem this feature will solve?

The SDMX proxy keeps its registry and agency routing configuration in a separate service, the [SDMX proxy config server](https://github.com/epam/statgpt-sdmx-proxy/tree/development/sdmx-proxy-config-server). That configuration decides which upstream registry serves each agency, so it is the thing an operator has to change when onboarding a new agency or repointing an existing one.

Today it is unreachable from StatGPT. Changing it means editing the object in DIAL Storage directly, or redeploying the config server with `CONFIG_SERVER_FORCE_SEED=true` to reseed from its bundled default — which also wipes every environment-specific change. There is no way for an admin to even see the configuration a proxy data source is running against.

## What is the proposed feature or solution?

Surface the configuration in the `details` field of proxy (`PROXY_SDMX30`) data sources, so the Admin UI can display and edit it with no new endpoints:

- **Read.** `GET /data-sources` and `GET /data-sources/{id}` populate `details.proxyConfig` from the config server.
- **Write.** The create and update endpoints send a submitted `details.proxyConfig` back to the config server.

The config server stays the single source of truth: the value is fetched live on every read and is never persisted in the `data_sources.details` column, so it cannot drift from what the proxy actually uses. Omitting the field means "leave the config server alone", which keeps a data source editable while its config server is unreachable.

The endpoint to use is already configurable per data source via `configUrl`, defaulting to `$env:{SDMX_PROXY_CONFIG_SERVER_HOST}/statgpt/sdmx-proxy-config-server/api/v0/config`.

A configuration rejected by the config server is reported to the admin; any other config-server failure is logged and leaves the data source readable, so one unreachable config server cannot break the data sources page.

The configuration should also travel with channel export/import and the CLI content sync, so a proxy setup can be moved between environments alongside the data sources that use it.

## What alternatives have you considered?

- Storing a copy of the configuration in `data_sources.details` and syncing it to the config server. Rejected: the proxy, a forced reseed, and direct storage edits can all change the configuration out-of-band, so the copy would silently drift and the Admin UI would show a stale configuration.

Fedir-Yatsenko:
## Applicable issues

- closes #569

## Description of changes

The SDMX proxy keeps its registry and agency routing configuration in a separate service, the SDMX proxy config server, which until now was unreachable from StatGPT: changing it meant editing the object in DIAL Storage by hand, or redeploying with `CONFIG_SERVER_FORCE_SEED=true`, which discards every environment-specific change.

This exposes it as `details.proxyConfig` on `PROXY_SDMX30` data sources, so the Admin Portal can show and edit it without new endpoints: the `GET` endpoints populate the field from the config server, and the create and update endpoints send a submitted value back to it. The config server remains the single source of truth — the value is read live and never persisted in `data_sources.details`, so it cannot drift from what the proxy actually uses. Omitting the field means "leave the config server alone", which keeps a data source editable while its config server is unreachable.

A configuration the config server rejects is reported to the admin as `422`; every other config-server failure is logged at `WARNING` and leaves the field absent, so one unreachable config server cannot break the data sources page.

The configuration also travels with channel export/import and the CLI content sync, so a proxy setup can be moved between environments. Change detection compares it only when the incoming config file specifies it, so a file that omits `proxyConfig` is not reported as a change on every run.

`DataSourceConfig` gains `dump_for_storage()` and `matches_stored()` so a type can declare which of its fields belong to an external system rather than to the database. Behaviour is unchanged for the other data source types.

## Deployment changes

New environment variable, read by the admin backend:

| Variable                        | Required                                           | Description                                                                                         |
|---------------------------------|----------------------------------------------------|-----------------------------------------------------------------------------------------------------|
| `SDMX_PROXY_CONFIG_SERVER_HOST` | Yes, if a `PROXY_SDMX30` data source is configured | Scheme, host and port of the SDMX proxy config server, e.g. `http://sdmx-proxy-config-server:8060`. |

It fills the default value of the data source's `configUrl` field, `$env:{SDMX_PROXY_CONFIG_SERVER_HOST}/statgpt/sdmx-proxy-config-server/api/v0/config`. When it is not set, the placeholder stays unresolved, no request is attempted, and the configuration is reported as unavailable — the data source itself keeps working. A deployment may instead set `configUrl` explicitly per data source.

No database migration and no change for other data source types.

## Checklist


Fedir-Yatsenko:
/deploy-review

>GitHub actions run: [30894861426](https://github.com/epam/ai-dial-ci/actions/runs/30894861426)

| [deploy-review](https://chat-statgpt-backend-pr-570.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/30894861426/job/91945246850) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/30894861426/job/91946811607) |

kryachkow (approved):

Issue #575, opened 2026-08-04
Deep Research: mediate the clarification flow through the Supreme Agent context


kryachkow:
> Sub-issue of #490. Second half of the #544 split; builds on the MVP tool set + button + `state` from #574.

### StatGPT Backend version

latest

### What is the problem this feature will solve?

Sub-issue of #490 (Add Deep Research tool).

The MVP (#574) makes the clarification flow work by passing messages through **verbatim**: whatever Deep Research asks is shown to the user as-is, and whatever the user replies goes straight back. That is fine to unblock testing, but it is not the target behaviour — the Supreme Agent must stay the single owner of the chat and its `state` and **mediate** the clarification flow instead of blindly forwarding.

Concretely, Deep Research over-asks: it requests clarifications the conversation already answers (e.g. a time range the user already gave). Forwarding those verbatim makes the user re-answer things needlessly. The Supreme Agent should absorb what it can from context and only surface the questions it genuinely cannot answer.

### What is the proposed feature or solution?

**1. Return Deep Research clarifications to the Supreme Agent as a tool response**
- When Deep Research returns a clarification, we do **not** parse or interpret it in code. We hand it back to the Supreme Agent as a normal tool response and let the agent decide what to do — answer it, or ask the user.
- Deep Research stays an **abstraction**: we keep as little knowledge of its internals as possible and avoid hardcoding/branching on the specific shape of the questions it sends.

**2. Answer from context, surface only what's needed**
- The Supreme Agent uses the conversation context to answer clarifications it can (e.g. a time range already provided) and, when Deep Research asks several questions at once, answers the ones it knows and surfaces only the remainder to the user.
- The routing between "continue the Deep Research session" and "this is a fresh Supreme Agent request" is driven by the session flag from #574.

**3. No extra summarization / hand-written LLM calls**
- Do not add a separate summarization step or bespoke prompt over the Deep Research exchange. Let the Supreme Agent's own tool-calling handle it — hand-written summarization has historically produced worse results than letting the model drive the calls.
- Keep the Deep Research ↔ Supreme Agent conversation separate from the user-facing conversation in `state`, so intermediate Deep Research history does not pollute the agent context.

> Analogy: Claude sub-agents' "send message" pattern — a sub-agent that has a question returns it to the main conversation, and the main agent resumes it via a follow-up message. We want the Deep Research resume flow to behave the same way.

### Edge cases to discuss

[...]

PR #576, opened 2026-08-05
feat: add Deep Research force mode with multi-turn clarification flow #574

### LINKED ISSUE #574: Deep Research: force button + MVP two-tool flow with verbatim clarification passthrough
> Sub-issue of #490. Splits #544 into two: this issue is the MVP that unblocks testers; the mediated clarification flow is tracked separately in #575.

### StatGPT Backend version

latest

Sub-issue of #490 (Add Deep Research tool).

The tool basis (#490, "basis" sub-issue) is single-shot: the Supreme Agent calls Deep Research with one `query` and streams the report back. Two things are missing before anyone can actually use it:

1. **No activation control.** Deep Research must be activated exclusively by the user clicking the dedicated UI control; today the Supreme Agent ignores any incoming `tool_choice` / config flag, so the button does nothing.
2. **No multi-turn support.** Deep Research is interactive — it asks clarifying questions and asks for plan approval before running — but we cannot carry that conversation across several user messages.

QA and stakeholders are already asking whether the button works. Rather than ship the button alone (which cannot complete a real Deep Research run) or wait for the full agent-mediated clarification flow, this issue delivers a **minimal end-to-end MVP**: the button forces the tool, and the clarification flow works by **passing messages through verbatim** in both directions. The Supreme Agent does not yet mediate — that is the follow-up.

**1. Button / forced activation**
- Honour the dedicated Deep Research UI control: when set, force the Supreme Agent to call the Deep Research start tool via `tool_choice` (or the `custom_fields.configuration.deep_research` flag we already read). The agent still composes the `query` argument itself.
- When the control is disabled/off for a channel, the tool is fully unavailable to the Supreme Agent.
- No more than one Deep Research session may run in a single chat at a time; a second concurrent run surfaces an error.

**2. Split the single tool into two**
- `deep_research(query)` — **starts** a session.
- `resume_deep_research(message)` — **resumes** the active session by sending the next message (answer to a clarification / plan approval).
- Splitting `start` from `resume` keeps each tool's system description clean and avoids one tool having to both kick off and continue a conversation.

**3. Persist the session in `state`**
- Store the Supreme Agent ↔ Deep Research exchange in DIAL `state` and track a session flag: `in_progress` vs `completed`.
- On a resume call, only the **new** message is sent as the tool argument; the rest of the Deep Research conversation is reconstructed from `state`. Keep the Deep Research conversation separate from the user-facing conversation so its intermediate history does not pollute the Supreme Agent context.

**4. Verbatim clarification passthrough (the MVP simplification)**
- While a session is `in_progress`, the next user message is routed to `resume_deep_research` and passed into Deep Research **verbatim** — no interpretation, no answering-from-context, no summarization.
- Whatever Deep Research returns (clarifying questions, plan-for-approval, final report) is surfaced to the user **verbatim**, reusing the existing streaming of `content` / `stages` / `attachments`.
- When the session is `completed`, control returns to the Supreme Agent and subsequent messages are handled normally.

**Out of scope (follow-up: #575):**
- Supreme Agent mediating clarifications — answering some questions itself from conversation context and only surfacing the ones it genuinely needs.

- **Ship the button only, without the two-tool flow.** Rejected: the button alone cannot complete a real Deep Research run (no clarification handling), so testers could only see the initial request — not enough to validate the feature.
- **Jump straight to the agent-mediated flow.** Deferred: it carries nuances (which questions the agent can answer, how much it absorbs vs passes through) that need real usage to tune. Verbatim passthrough is a smaller, shippable step that unblocks testing now.

kryachkow:
## Applicable issues

- closes #574

## Description of changes

Adds a "force Deep Research" mode driven by a single `deep_research` toggle the frontend sends on each request. When the toggle is on, the Supreme Agent forces the Deep Research start tool and drives a two-tool flow that carries Deep Research's multi-turn clarification / plan-approval exchange across user turns.

- **Toggle-gated routing.** `deep_research` is the single switch: when on, a new request forces the start tool as the only bound tool (`tool_choice`, with parallel tool calls disabled so the model cannot emit more than one Deep Research call per response) and an in-progress session is resumed; when off, the tool is hidden from the agent and any in-progress session is abandoned. The toggle is advertised to the frontend only where the channel has the Deep Research tool enabled.
- **Two-tool flow, built on demand.** The start tool is built on demand from its channel config, and a `resume_deep_research` tool is derived from the same config to continue an in-progress session. Because Deep Research is deliberately excluded from the config-scoped executor (and the MCP server), the forced call is dispatched through a `ToolCaller` that knows the on-demand tool. The resume tool's name and description are admin-configurable via `DeepResearchDetails.resume_tool_name` / `resume_tool_description`, and the run logic lives in a dedicated `DeepResearchRunner`.
- **Verbatim passthrough.** While a session is in progress, the user's next message is forwarded to Deep Research exactly as written (bypassing the LLM), and its reply — clarifying question, plan for approval, or final report — is streamed back as-is.
- **Session persistence.** A `DeepResearchSession` in DIAL state stores the sub-conversation as typed `DeepResearchTurn` records (user message, assistant content, and Deep Research's own `deep_research_state`) rather than DIAL-message-shaped objects: the DIAL chat client strips `custom_content` from message-shaped objects while round-tripping state, which would drop `deep_research_state` and force a re-plan every turn. The DIAL messages are rebuilt from these fields only when calling the deployment. The session's mere presence signals an in-progress run and is dropped once the final report is delivered, so state stays small (the report already lives in chat history).
- **Unified failure handling.** The tool no longer surfaces its own errors: request/stream failures (and any other exception) propagate out of `DeepResearchRunner.run` to `ToolCaller.call_tool`, which records the error in the tool state as an ERROR tool message, exactly like any other tool failure. Both the forced-start path (`SupremeAgentExecutor.stream_response`) and the resume path (`_resume_deep_research`) then funnel that ERROR through a single `surface_deep_research_error` helper, so the standard failure message is streamed to the user exactly once and the wording lives in one place.
- Unit tests cover the session/turn helpers, the forced-start dispatch and session round-trip, the resume/verbatim passthrough, the failure path (error recorded in the tool state and the message surfaced exactly once), and the streamer's state capture.

Verbatim passthrough is the MVP behavior; mediated clarification (the agent answering from context instead of forwarding) is the follow-up in #575.

## Checklist


statgpt/app/chains/deep_research/deep_research_tool.py
```diff
@@ -1,29 +1,50 @@
...
+)
 from statgpt.app.utils import OpenAiToDialStreamer, openai
 from statgpt.common.config import multiline_logger as logger
+from statgpt.common.schemas import ChannelConfig
 from statgpt.common.schemas import DeepResearchTool as DeepResearchToolConfig
 from statgpt.common.schemas import ToolTypes
 from statgpt.common.schemas.llm_call_duration import LLMCallDurationItem
 from statgpt.common.utils.llm_call_duration_context import get_llm_call_duration_manager
 
+RESUME_DEEP_RESEARCH_TOOL_NAME = "resume_deep_research"
```
Fedir-Yatsenko: Let's add this to the Deep Research config

statgpt/app/chains/deep_research/deep_research_tool.py
```diff
@@ -33,65 +54,180 @@ def get_args_schema(cls, tool_config: DeepResearchToolConfig) -> type[DeepResear
...
         deployment_id = details.get_deployment_id()
 
-        client = openai.get_async_client(api_key=auth_context.api_key)
-
-        create_kwargs: dict[str, Any] = dict(
-            model=deployment_id,
-            stream=True,
-            messages=self._construct_history(query),
+        session = DeepResearchTool._load_session(state)
+        messages = DeepResearchTool._build_request_messages(
```
Fedir-Yatsenko: Why not use `cls.<func>`?

statgpt/app/chains/deep_research/deep_research_tool.py
```diff
@@ -33,65 +54,180 @@ def get_args_schema(cls, tool_config: DeepResearchToolConfig) -> type[DeepResear
...
+    @classmethod
+    def build(cls, tool_config: DeepResearchToolConfig, channel_config: ChannelConfig) -> Self:
+        """Construct the resume tool from the (start) Deep Research tool config. The name and
+        description are fixed rather than admin-configured, since this tool is an internal
+        continuation mechanism."""
+        return cls(
+            tool_config=tool_config,
+            channel_config=channel_config,
+            name=RESUME_DEEP_RESEARCH_TOOL_NAME,
+            description=RESUME_DEEP_RESEARCH_TOOL_DESCRIPTION,
```
Fedir-Yatsenko: `name` and `description` must be configurable, let's add them to the Deep Research config

statgpt/app/chains/supreme_agent.py
```diff
@@ -274,13 +286,47 @@ async def stream_response(self, inputs: dict) -> str:
...
+            if configuration.deep_research:
+                if session_in_progress:
+                    # Active session + toggle on: forward the user's next message to Deep Research
+                    # verbatim (no interpretation) and surface its reply as-is. Deterministic
+                    # routing driven by the session flag, so a resume never depends on the LLM
+                    # re-composing the message.
+                    return await self._resume_deep_research(inputs, deep_research_tool)
+                tool_choice = deep_research_tool.name
+            else:
+                agent_tools = [t for t in agent_tools if t.tool_type != ToolTypes.DEEP_RESEARCH]
```
Fedir-Yatsenko: Let's exclude Deep Research from the list of tools. It should only be accessible via a button.

statgpt/app/chains/supreme_agent.py
```diff
@@ -351,6 +423,81 @@ async def stream_response(self, inputs: dict) -> str:
...
+        }
+        tool_msg = await ToolCaller([resume_tool]).call_tool(tool_call, inputs, show_stage=False)
+        if tool_msg.status == ToolResponseStatus.ERROR.value:
+            choice.append_content(DEEP_RESEARCH_ERROR_MESSAGE)
+            return DEEP_RESEARCH_ERROR_MESSAGE
+        content = tool_msg.content
+        return content if isinstance(content, str) else str(content)
+
+    @staticmethod
+    def _last_user_message_text(history: History) -> str:
```
Fedir-Yatsenko: Let's make this a `History` property

statgpt/app/chains/supreme_agent.py
```diff
@@ -351,6 +423,81 @@ async def stream_response(self, inputs: dict) -> str:
...
+            if text := getattr(part, "text", None):
+                parts.append(text)
+        return "\n".join(parts)
+
+    @staticmethod
+    def _guard_deep_research_calls(
+        tool_calls: list[ToolCall], deep_research_tool: StatGptTool | None
+    ) -> tuple[list[ToolCall], list[ToolMessage]]:
+        """Enforce a single Deep Research run per agent response: if the agent emits more than
+        one Deep Research call in one response, keep the first and reject the rest with an error.
```
Fedir-Yatsenko: https://developers.openai.com/api/docs/guides/function-calling#parallel-function-calling

> The model may choose to call multiple functions in a single turn. You can prevent this by setting `parallel_tool_calls` to `false`, which ensures exactly zero or one tool is called.

statgpt/app/schemas/dial_app_configuration.py
```diff
@@ -28,6 +28,15 @@ class StatGPTConfiguration(BaseModel):
         "When False, each dataset produces a separate python code attachment.",
         default=False,
     )
+    # Keep in sync with `DeepResearchChannelConfiguration` in services/chat_facade.py, which
+    # advertises this toggle to the frontend only when the Deep Research tool is enabled. This
+    # model parses the value the frontend sends back on each request.
+    deep_research: bool = Field(
```
Fedir-Yatsenko, line 31: Why is this property needed?

statgpt/app/chains/deep_research/deep_research_tool.py
kryachkow: done

statgpt/app/chains/deep_research/deep_research_tool.py
kryachkow: done

statgpt/app/chains/supreme_agent.py
kryachkow: fixed

statgpt/app/chains/supreme_agent.py
kryachkow: done

statgpt/app/chains/supreme_agent.py
kryachkow: fixed

statgpt/app/schemas/dial_app_configuration.py
kryachkow: It's the typed parse target for the toggle the frontend sends back each request (configuration.deep_research drives routing); extra='allow' would capture it untyped, but declaring it gives a typed bool and better readability

statgpt/app/chains/deep_research/deep_research_tool.py
```diff
@@ -23,75 +30,231 @@ class DeepResearchArgs(ToolArgs):
...
+                    stream = await client.chat.completions.create(**create_kwargs)
                     async for chunk in stream:
                         dial_streamer.send_chunk(chunk)
-                except APIError as e:
+                except OpenAIError as e:
+                    # Catch deployment-call failures (connection/timeout/status/stream errors)
+                    # and surface a friendly message; let programming errors propagate.
                     logger.exception(e)
-                    response = "<error>Something went wrong</error>"
+                    failed = True
```
Fedir-Yatsenko: Let's put the error in the `state` as well

statgpt/app/chains/supreme_agent.py
```diff
@@ -328,6 +409,16 @@ async def stream_response(self, inputs: dict) -> str:
...
                     )
                     return ""
+
+                if deep_research_msg is not None:
+                    # Deep Research owns this turn: it streamed its clarifying question / final
+                    # report straight into the conversation, or failed. Either way, end the loop —
+                    # under the forced tool_choice, continuing would just re-invoke Deep Research
+                    # and, on a persistent failure, spin until max_agent_iterations.
+                    if deep_research_msg.status == ToolResponseStatus.ERROR.value:
+                        return surface_deep_research_error(choice)
```
Fedir-Yatsenko, line 401: It looks like this function is being called twice.

statgpt/app/chains/deep_research/deep_research_tool.py
kryachkow: fixed

statgpt/app/chains/supreme_agent.py
kryachkow: resolved alongside the "error in state" comment. DeepResearchRunner.run no longer surfaces errors itself: request/stream failures now propagate to ToolCaller.call_tool, which records them as an ERROR tool message. This line is now the single surface point for the forced-start path (and _resume_deep_research for the resume path), so surface_deep_research_error fires exactly once per failure.

kryachkow:
/deploy-review

>GitHub actions run: [32248405869](https://github.com/epam/ai-dial-ci/actions/runs/32248405869)

| [deploy-review](https://chat-statgpt-backend-pr-576.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/32248405869/job/96053851725) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/32248405869/job/96055821425) |

Fedir-Yatsenko (approved):

PR #584, opened 2026-08-11
fix: Session-level advisory locks are leaked onto pooled connections

### LINKED ISSUE #583: Session-level advisory locks are leaked onto pooled connections
### StatGPT Backend version

0.13.2

### Short description

Dataset advisory locks are taken with session-level `pg_try_advisory_lock`, but the surrounding code commits while holding them. SQLAlchemy releases the connection back to the pool on `commit()`, so the matching `pg_advisory_unlock` runs on a *different* connection, silently fails, and the lock stays held by an idle pooled connection. Subsequent reindex / clear operations on the same (collection, dataset) block for the full 600 s timeout and fail with `TimeoutError: Could not acquire advisory lock ...`.

1. Attach a dataset to a channel and run indexing for it (any dataset with at least one non-indicator and one
   special dimension.
2. Let it complete successfully. The log shows `Added N documents` / `Added N document mappings` for both
   `AvailableDimensions_<ch>` and `SpecialDimensions_<ch>`.
3. Trigger any further reindex or clear of the same dataset.
4. It hangs for `advisory_lock_timeout` (default 600 s,
   [`settings/database.py:45`](https://github.com/epam/statgpt-backend/blob/e37f579c744f4501964f5888a9c56841c4d39906/statgpt/common/settings/database.py#L45))
   and fails.

Step 2 is the one that plants the fault — a **successful** write leaks the lock. Nothing needs to run
concurrently; the second operation can be minutes or hours later, on an otherwise idle process.

`_dataset_lock` should guarantee mutual exclusion per (collection, dataset) for the duration of the guarded
block, and release the lock when the block exits — so a completed reindex leaves no lock behind and the next
operation proceeds immediately.

### What do you see instead?

```
ERROR:    | 2026-08-07 15:05:00 | 6 | statgpt.admin.services.dataset | Failed to reindex version_id=433
Traceback (most recent call last):
  File "/home/app/statgpt/admin/services/dataset.py", line 1945, in reload_channel_dataset_in_background
    await self._index_non_indicator_dimensions(
  File "/home/app/statgpt/admin/services/dataset.py", line 1787, in _index_non_indicator_dimensions
    await vector_store.add_documents(documents, dataset_id=dataset.id, version_id=version_id)
  File "/home/app/statgpt/common/vectorstore/pg_vector_store/pg_vector_store.py", line 620, in add_documents
    async with self._dataset_lock(session, dataset_id):
  File "/home/app/statgpt/common/vectorstore/pg_vector_store/pg_vector_store.py", line 239, in _dataset_lock
    lock_key = await self._acquire_dataset_lock(session, dataset_id)
  File "/home/app/statgpt/common/vectorstore/pg_vector_store/pg_vector_store.py", line 205, in _acquire_dataset_lock
    raise TimeoutError(
TimeoutError: Could not acquire advisory lock for collection=SpecialDimensions_12 dataset=dee4481b-33d9-5268-aeab-decb5f071821 (key=-8243928484966154159) within 600.0s
```

### Additional information

**The mutual exclusion never actually held.** Beyond the timeouts, the lock is dropped at the first `commit()`
inside the guarded block, so concurrent writers to the same (collection, dataset) were never excluded. Any
correctness argument that rests on `_dataset_lock` is currently unsupported — this is the quieter half of the
defect.

**Second leak, on the failure path.**
[`_dataset_locks`](https://github.com/epam/statgpt-backend/blob/e37f579c744f4501964f5888a9c56841c4d39906/statgpt/common/vectorstore/pg_vector_store/pg_vector_store.py#L245-L261)
acquires every lock in a
[list comprehension placed *outside* the `try`](https://github.com/epam/statgpt-backend/blob/e37f579c744f4501964f5888a9c56841c4d39906/statgpt/common/vectorstore/pg_vector_store/pg_vector_store.py#L253-L261).
If the second acquisition times out, the first is never released. The 15:15:01 traceback is raised from inside
that comprehension. Not the trigger here (a single dataset id was locked), but it compounds the same failure.

**Why it looks intermittent.** `pool_recycle=1800` retires a leaked connection at its next checkout, so locks
free themselves after roughly half an hour of traffic — and a leaked connection later reused by another task
re-acquires the lock re-entrantly (counter 2), whose unlock brings it back to 1, making that call appear to
succeed. Hence the same key failing repeatedly over hours with sporadic successes in between.

iliia-veselov:
## Applicable issues

- fixes #583 

## Description of changes

Fixed leak of session-level locks on pooled connections.

## Checklist

### DIFF
--- a/statgpt/common/models/__init__.py
+++ b/statgpt/common/models/__init__.py
@@ -1,5 +1,6 @@
 from .database import (
     SessionMakerSingleton,
+    get_engine,
     get_readonly_session_context_manager,
     get_session,
     get_session_context_manager,
--- a/statgpt/common/models/database.py
+++ b/statgpt/common/models/database.py
@@ -100,6 +100,12 @@ class SessionMaker:
     def __init__(self, engine_config: dict):
         self._engine_config = engine_config
         self._postgres_settings = PostgresSettings()
+        self._engine: AsyncEngine | None = None
+
+    @property
+    def engine(self) -> AsyncEngine | None:
+        """The engine created by `create()`, or None if `create()` has not been called."""
+        return self._engine
 
     @staticmethod
     async def _test_connection(engine: AsyncEngine) -> bool:
@@ -190,6 +196,7 @@ class SessionMaker:
     async def create(self) -> async_sessionmaker[AsyncSession]:
         _log.debug("Creating session maker")
         engine = await self.create_engine()
+        self._engine = engine
         session_maker = async_sessionmaker(
             engine,
             expire_on_commit=False,
@@ -202,6 +209,7 @@ class SessionMaker:
 
 class SessionMakerSingleton:
     instance: async_sessionmaker[AsyncSession] | None = None
+    engine: AsyncEngine | None = None
     _lock = asyncio.Lock()
 
     @classmethod
@@ -217,7 +225,9 @@ class SessionMakerSingleton:
                 return cls.instance
 
             _log.debug("Creating new SessionMakerSingleton instance")
-            cls.instance = await SessionMaker(SessionMaker.DEFAULT_ENGINE_CONFIG).create()
+            session_maker = SessionMaker(SessionMaker.DEFAULT_ENGINE_CONFIG)
+            cls.instance = await session_maker.create()
+            cls.engine = session_maker.engine
             return cls.instance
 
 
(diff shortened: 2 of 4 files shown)


Fedir-Yatsenko:
/deploy-review

>GitHub actions run: [31495646366](https://github.com/epam/ai-dial-ci/actions/runs/31495646366)

| [deploy-review](https://chat-statgpt-backend-pr-584.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/31495646366/job/93792546153) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/31495646366/job/93795247322) |

Fedir-Yatsenko:
Rethought in a separate pull request (#591).

PR #588, opened 2026-08-11
fix: SDMX TimeRange in an availability response aborts the whole dataset

### LINKED ISSUE #587: SDMX `TimeRange` in an availability response aborts the whole dataset
### StatGPT Backend version

0.13.2

### Short description

OECD expresses `TIME_PERIOD` availability with the standard SDMX 2.1 `<common:TimeRange>` element. StatGPT assumes every cube-region member exposes `.value`, so the conversion raises `AttributeError`. Because that conversion is a single funnel for the whole cube region, one unreadable time member takes down **every** dimension of the dataset — it can then be neither indexed nor queried.

**Case 1 — the raw conversion (no StatGPT setup needed)**

1. Fetch a real availability response that uses a time range:
   ```bash
   curl -o oecd.xml \
     'https://sdmx.oecd.org/public/rest/availableconstraint/OECD.SDD.TPS,DSD_PRICES@DF_PRICES_ALL,1.0/all/all'
   ```
2. Run the exact comprehension from `statgpt/common/data/sdmx/v21/dataset.py:1108`:
   ```python
   import sdmx

   message = sdmx.read_sdmx("oecd.xml")
   cube_region = list(message.constraint.values())[0].data_content_region[0]
   {dim.id: {v.value for v in sel.values} for dim, sel in cube_region.member.items()}
   ```
3. It raises immediately — see "What do you see instead?".

**Case 2 — onboarding (the reported symptom)**

1. Configure an SDMX 2.1 data source pointing at `https://sdmx.oecd.org/public/rest`.
2. Add the dataflow `OECD.SDD.TPS:DSD_PRICES@DF_PRICES_ALL(1.0)` to a channel.
3. Trigger indexing (`channel reindex`, or the Admin Portal's reindex action).
4. Indicator enumeration stops on the first availability call. The dataset ends with no
   indicator embeddings, so it is also unanswerable in chat.

* `<common:TimeRange>` is a valid SDMX 2.1 availability construct and must be understood, not
  crashed on. Its bounds belong in `DataSetAvailabilityQuery.time_period_start` /
  `time_period_end` — fields that already exist and are already consumed by the chat backend to
  check whether a requested period overlaps what the dataset holds.
* A member value type StatGPT does not recognise must be skipped with a warning, leaving the
  rest of the cube region intact. Losing one dimension's detail is a degradation; losing the
  dataset is an outage.
* The dataset indexes and answers queries like any other.

### What do you see instead?

```
AttributeError: 'RangePeriod' object has no attribute 'value'
```

`sdmx1` parses `<common:Value>` into `MemberValue` (which has `.value`) but `<common:TimeRange>`
into `RangePeriod` (which has `.start` / `.end` instead). `<common:BeforePeriod>` and
`<common:AfterPeriod>` parse into `BeforePeriod` / `AfterPeriod`, which have `.period`. None of
these are handled.

The OECD response mixes both kinds in one cube region — eight dimensions of `<common:Value>`
plus one `TIME_PERIOD` time range:

```xml
<common:KeyValue id="TRANSFORMATION">
  <common:Value>G1</common:Value>
  <common:Value>GOY</common:Value>
</common:KeyValue>
<common:KeyValue id="TIME_PERIOD">
  <common:TimeRange>
    <common:StartPeriod isInclusive="true">1914-01-01T00:00:00</common:StartPeriod>
    <common:EndPeriod   isInclusive="true">2026-07-31T00:00:00</common:EndPeriod>
  </common:TimeRange>
</common:KeyValue>
```

Two call sites make the assumption:

| Location | Code | Reached from |
|---|---|---|
| `statgpt/common/data/sdmx/v21/dataset.py:1108` | `{v.value for v in selection.values}` | every `availability_query()` call — **the reported crash** |
| `statgpt/common/data/sdmx/v21/dimensions_creator.py:237` | `set([item.value for item in member_dict[dimension.id].values])` | dataflow load; only queried for non-time code-list dimensions today, so latent |

**Blast radius.** `_availability_result_to_query` is the single conversion point behind
`Sdmx21DataSet.availability_query()` (`dataset.py:1173`), so the one `AttributeError` breaks
both halves of the product:

* **Indexing** — `_load_indicator_combinations_to` (`dataset.py:627`) and the recursive
  `_get_available_series` (`dataset.py:545`). This is the "indicator-enumeration process
  stopped" symptom.
* **Answering** — `query_builder/dimensions/base.py:61`,
  `indicator_selection/semantic/indicators_semantic_v1.py:195`,
  `query_builder/query/finalize_query.py:182`, `query_constructor/composite.py:70`,
  `query_constructor/iterative.py:245`.

### Additional information

_No response_

iliia-veselov:
## Applicable issues

- fixes #587 

## Description of changes

Reading a cube-region member selection now dispatches per value type instead of assuming every member is a coded `<common:Value>`: `read_member_selection` walks a selection once and returns its coded values alongside the envelope of its time ranges, skipping anything it cannot interpret with a warning, so an unsupported value type can no longer abort the whole conversion. The range is not just survived but used — a `<common:TimeRange>` on the configured time period dimension populates `time_period_start` / `time_period_end` as `YYYY-MM-DD`, matching what the time-period checks in `finalize_query` already compare against, while a range on any other dimension is ignored and logged rather than allowed to widen the dataset's availability.

Note this means bounds are now populated for **all** SDMX 2.1 providers rather than only QuantHub, so a request lying entirely outside availability can now be reported as `INVALID_TIME_PERIOD` where previously there was nothing to check it against.

## Checklist


statgpt/common/data/quanthub/v21/dataset.py
```diff
@@ -96,8 +102,13 @@ def _parse_time_period_from(annotations: list[BaseAnnotation]) -> tuple[str | No
...
             elif annotation.id == "time_period_end":
-                end = annotation.value or annotation.title
+                end = _annotation_value(annotation)
 
         return start, end
+
+
+def _annotation_value(annotation: BaseAnnotation) -> str | None:
+    # `sdmx1` annotations carry the payload in `title`; SDMX 3.0 ones in `value`.
+    return getattr(annotation, "value", None) or annotation.title
```
Fedir-Yatsenko: `sdmx1` implements the `value` property for both SDMX 2.1 and SDMX 3.0. See the [source code](https://github.com/khaeru/sdmx/blob/1bbb0c757fc417620f444a3dd767abd0759f7188/sdmx/model/common.py#L174). So I wonder, do we need these changes?

Fedir-Yatsenko:
/deploy-review

>GitHub actions run: [31714206982](https://github.com/epam/ai-dial-ci/actions/runs/31714206982)

| [deploy-review](https://chat-statgpt-backend-pr-588.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/31714206982/job/94494639506) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/31714206982/job/94496588264) |

Fedir-Yatsenko (dismissed):

statgpt/common/data/sdmx/v21/dimensions_creator.py
```diff
@@ -236,4 +237,6 @@ def _get_available_dimension_values(self, dimension: common.DimensionComponent)
         member_dict = cube_region.member
         if dimension.id not in member_dict.keys():
             raise ValueError(f"Missing dimension({dimension.id}) value in data content constraint")
-        return set([item.value for item in member_dict[dimension.id].values])  # type: ignore
+        # This is only called for non-time code-list dimensions, so a time range here would be a
+        # provider quirk: `member_value_ids` warns and returns an empty set instead of raising.
+        return member_value_ids(member_dict[dimension.id])  # type: ignore[index]
```
Fedir-Yatsenko: **The comment above comment describes behavior the code does not have:**

`member_value_ids` does **not** warn for a time range — `RangePeriod`/`AfterPeriod`/`BeforePeriod` are explicitly excluded from the `unsupported` set (`member_selection.py:32`), so they are skipped silently.

statgpt/common/data/sdmx/v21/member_selection.py
```diff
@@ -0,0 +1,86 @@
...
+            start = min_bound(start, value.start.period.strftime(DATE_FORMAT))
+            end = max_bound(end, value.end.period.strftime(DATE_FORMAT))
+        elif isinstance(value, AfterPeriod):
+            start = min_bound(start, value.period.strftime(DATE_FORMAT))
+        elif isinstance(value, BeforePeriod):
+            end = max_bound(end, value.period.strftime(DATE_FORMAT))
+        elif not isinstance(value, MemberValue):
+            unsupported.add(type(value).__name__)
+
+    _log_unsupported(selection, unsupported)
```
Fedir-Yatsenko: **Unsupported value types are warned about twice per dimension**

`statgpt/common/data/sdmx/v21/member_selection.py:34,65` — both `member_value_ids` and `time_range_bounds` end with `_log_unsupported`, and `dataset.py:1111-1112` calls both for every dimension. A genuinely unknown `SelectionValue` therefore produces two identical warnings per dimension per availability response.

statgpt/common/data/sdmx/v21/dataset.py
```diff
@@ -1104,18 +1105,26 @@ def _availability_result_to_query(
...
-            result.add_dimension_query(
-                DimensionQuery(
-                    dimension_id=dimension_id,
-                    values=list(available_values),
-                    operator=QueryOperator.IN,
+
+        result = DataSetAvailabilityQuery()
+        for dim, selection in member_dict.items():
+            coded_values = member_value_ids(selection)
+            range_start, range_end = time_range_bounds(selection)
```
Fedir-Yatsenko: **Time bounds are harvested from *every* dimension, not just the time dimension**

`dataset.py:1110-1117` calls `time_range_bounds` for each dimension and folds any result into `result.time_period_start/end`. A range on a non-time dimension — the same "provider quirk" that `dimensions_creator.py` warns about — would silently widen the dataset's time bounds.

`Sdmx21DataSet` knows its time dimension (`get_time_dimension()` / `config.time_period_dimension`), so the extraction could be scoped to it.

Fedir-Yatsenko:
/deploy-review

>GitHub actions run: [32016809088](https://github.com/epam/ai-dial-ci/actions/runs/32016809088)

| [deploy-review](https://chat-statgpt-backend-pr-588.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/32016809088/job/95347914139) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/32016809088/job/95349481937) |

Fedir-Yatsenko (approved):

PR #643, opened 2026-09-04
feat: expose stable record identifiers in MCP tool results #608

### LINKED ISSUE #608: MCP app: Expose stable record identifiers in MCP tool results
### StatGPT Backend version
latest

Follow-up tool calls need to reference records returned by an earlier call. Without stable identifiers the model has to re-identify records by label, which is ambiguous across similarly named datasets, indicators and series, and forces redundant queries.

StatGPT results do not consistently carry an identifier that is safe to pass back in a later call.

Required by: [OpenAI: Build MCP server](https://developers.openai.com/plugins/build/mcp-server).

**Assign an ID to every returned entity.** Datasets, dataflows, indicators, series and observations each carry an identifier in `structuredContent`.

**Make IDs stable.** The same logical record keeps the same identifier across sessions and releases; document the composition rules (for example provider + dataflow + key) and what would constitute a breaking change.

**Accept IDs as inputs.** Follow-up tools take those identifiers directly as parameters, so a second call can act on the exact record the first call returned.

**Document the format.** Describe the identifier format in the tool schemas so the model knows it is opaque and must be passed through verbatim.

**Definition of done.** A round-trip test performs a search, feeds returned identifiers into the detail and observation tools, and asserts the same records come back.

- **Match by name or label** — inadequate because names are not unique across providers and change between vintages.
- **Session-scoped ephemeral IDs** — insufficient since they break as soon as the conversation is resumed or the server scales.
- **Return full record payloads instead of IDs** — rejected: it multiplies payload size and conflicts with the result-size budget.

kryachkow:
## Applicable issues

- closes #608

## Description of changes

Each queried record in the MCP data query result now carries a stable, opaque `recordId`, so a follow-up call can reference the exact same SDMX record an earlier call returned without re-identifying it by label. Labels are not unique across providers and change between vintages, which made them unsafe to pass back.

The identifier is `AGENCY:RESOURCE(VERSION)/SERIES_KEY`: the dataflow short URN plus the SDMX 2.1 REST series key in DSD order. Because it is a pure function of the dataflow reference and series key, the same logical record keeps the same id across sessions and server instances; it only changes when the dataflow version or DSD layout changes.

- Add `common/schemas/record_id.py` with compose/parse/round-trip helpers, and move series-key building here from the python code generator.
- Add the `recordId` computed field to `AppJsonQueryWithMetadata`, and document the dataflow `urn` and dataset `id` as opaque, pass-back-verbatim values.
- Extend the SDMX proxy tool description to explain building request paths from the returned `recordId` / `urn`.
- Bump the structured content schema version 2 -> 3.

## Checklist


Fedir-Yatsenko:
Let's discuss the problem statement.

PR #645, opened 2026-09-07
feat: lint mcpDescription directly and require it for model-visible MCP tools #610

### LINKED ISSUE #610: MCP app: Ensure tool descriptions and annotations don't manipulate model tool selection
### StatGPT Backend version
latest

Both marketplaces prohibit using descriptions or annotations to manipulate tool selection or to interfere with fair discovery. In practice this covers priority claims ("always call this first"), keyword stuffing, disparaging other providers, and safety annotations that understate what a tool does.

This overlaps with the description-accuracy and coercion tasks but needs its own gate, because the failure mode is stylistic and easy to reintroduce in a routine PR.

Required by: [OpenAI: Submission guidelines](https://developers.openai.com/plugins/app-guidelines#fair-play)

**Define a ban list.** Priority/exclusivity claims, superlatives about other services, keyword stuffing, invisible or encoded instructions, and any text addressed to the model's selection logic rather than to a human reader.

**Lint tool metadata in CI.** A check that scans descriptions and annotations against the ban list and fails the build on a hit.

**Honest annotations.** Verify `readOnlyHint`, `destructiveHint` and `openWorldHint` reflect real behaviour — mislabelled annotations are a top rejection cause and also read as selection manipulation.

**Human gate.** Require a reviewer sign-off on any change to tool metadata, using the same checklist as the linter.

**Definition of done.** The linter runs in CI, the current tool set passes, and the annotation audit table is attached to the submission package.

- **Manual review only** — insufficient because the pattern is reintroduced incrementally by unrelated changes.
- **Trust the description-accuracy audit to cover it** — inadequate since an accurate description can still contain steering language.
- **Do nothing until a reviewer flags it** — rejected: a flagged submission restarts the review cycle.

kryachkow:
## Applicable issues

- part of #610

## Description of changes

Tightens the marketplace-policy linter so it validates the metadata that is actually published to MCP.

- `lint_tool` now lints `mcp_description` directly instead of `effective_mcp_description`. The name is still taken from `effective_mcp_name` (tool names are constrained identifiers that are safe to publish).
- A model-visible tool on an opted-in channel that declares no `mcp_description` is now a violation (`MISSING_MCP_DESCRIPTION`), rather than silently falling through to the agent-facing `description`.

**Why:** with a single channel serving both the Supreme Agent and MCP, tools keep steering language in `description` and a clean override in `mcp_description`. The old fallback only caught a forgotten override when the internal text happened to trip the ban list — a clean-looking but internal description would ship to MCP unchecked. Requiring an explicit override closes that hole and matches the repo's strict-contract / no-silent-fallback principle.

- Updated `test_mcp_policy.py` to the new contract: steering fixtures now live in `mcpDescription`, added `test_missing_mcp_description_is_flagged`, and removed a test that became a duplicate of the override case.

## Checklist

- [x] Title of the pull request follows the Conventional Commits specification


Fedir-Yatsenko:
Tools within a single plugin are allowed to contain instructions mandating that priority be given to other tools within that same plugin. For example, "use the Available Terms tool to obtain a list of valid arguments for the Get Term Definition tool". Even if it might not be the best approach, I wouldn't forbid it programmatically.

PR #657, opened 2026-09-09
feat: expose Grade C filter values, CSV export and import modes #651 #652 #653

### LINKED ISSUE #651: Grade C: expose the available filter values for a channel's discovery datasets
### StatGPT Backend version

latest

`GET /{channel_id}/discovery-datasets` accepts `validation_status`, `indexing_status` and `agency` as filters, but the Admin UI has no way to learn which values it may pass.

The two status filters are enums, so a generated client can enumerate them from the OpenAPI schema. `agency` is free text supplied by whoever filled in the discovery workbook: the set of agencies present in a channel is only known to the database. The UI is therefore left with a free-text input where a picker belongs — the admin has to know the exact agency name up front, and a typo returns an empty list rather than an error.

Expose the distinct agency values a channel actually holds, together with the record count for each, so the UI can render an agency picker instead of a text field.

Natural home is the existing `GET /{channel_id}/discovery-datasets/stats`, which already exists to let a caller render the whole breakdown from one request. Adding a `by_agency` map to `DiscoveryDatasetStats` keeps the filter metadata in one place and needs no new route.

Should the agency breakdown not belong in the stats payload (it is unbounded, unlike the status enums), the alternative is a dedicated `GET /{channel_id}/discovery-datasets/agencies` returning the distinct values with their counts.

Values should be returned as stored, and matched the way the `agency` filter already matches — ignoring case and surrounding whitespace — so that a value taken from the response is guaranteed to be a usable filter argument.

Once the breakdown is available, `statgpt-cli channel status` should render it next to the validation and indexing breakdowns it already prints.

Having the UI derive the agency list client-side by paging through every record of the channel. This works only for small channels, costs a full scan on every page load, and breaks as soon as the list is paginated by the server.

### LINKED ISSUE #652: Grade C: export a channel's discovery datasets as CSV
### StatGPT Backend version

latest

Discovery datasets can be loaded into a channel on their own — `POST /{channel_id}/discovery-datasets/upload` takes a workbook or a CSV — but they can only be got back out as part of a full channel export. Moving a set of records from one channel to another, or handing them to an agency for review, today means running an export job, downloading the whole zip and picking `discovery_datasets.csv` out of it.

That is a heavy round trip for one file, and it drags along the channel's configuration and datasets, which the recipient neither needs nor should necessarily see.

Add `GET /{channel_id}/discovery-datasets/export`, returning the channel's discovery datasets as a CSV download.

The file must be byte-for-byte the one the channel export writes — same columns, same order, produced by the same code path (`AdminPortalDiscoveryDatasetService.export_discovery_datasets_to_folder`) — so that what comes out of one channel goes straight into `POST /{channel_id}/discovery-datasets/upload` on another. Only the descriptive fields belong in it: validation and indexing state is per channel and is re-derived by the next indexing job.

Synchronous, unlike the channel export: a CSV of a channel's records is small enough to stream in the response, so it needs no background job to poll.

The endpoint must accept the list endpoint's `validation_status`, `indexing_status` and `agency` filters, with the same semantics, so that an admin can export exactly the slice they are looking at. With no filters supplied it exports the whole channel.

Keeping the channel export as the only way out, and documenting the "export, unzip, take one file" path. It works, but it makes a routine transfer disproportionately expensive and shares more of the channel than the task calls for.

### LINKED ISSUE #653: Grade C: support per-collection upsert/replace modes on channel import
### StatGPT Backend version

latest

`POST /{channel_id}/discovery-datasets/upload` takes a `mode`: `upsert` keeps the records the file does not mention, `replace` deletes them. Importing the same records as part of a channel archive gives no such choice — `POST /channels/import` merging into an existing channel always upserts, so records left over from an earlier import stay behind.

The result is that the two ways of loading the same CSV behave differently, and there is no way to make a channel's discovery datasets match an archive exactly other than clearing them first with `DELETE /{channel_id}/discovery-datasets/bulk`.

Add two query parameters to `POST /channels/import` — `discovery_datasets_mode` and `glossary_terms_mode` — reusing the `upsert` / `replace` semantics of the upload endpoint and defaulting to `upsert` so current behavior is unchanged.

Each parameter governs its own collection: an admin can replace the discovery datasets while leaving the channel's own glossary terms alone, or the other way round. `replace` makes the collection match the archive, including when the archive carries no file for it at all, which describes a channel with none.

Both only apply when merging into an existing channel: with `clean_up=true`, or when the import creates the channel, there is nothing to delete.

A single shared `mode` covering both collections. It is a smaller surface, but the two collections are curated independently — a channel's glossary is often maintained locally while its discovery datasets come from an archive — so one switch forces an admin to choose between leaving stale records behind and deleting terms they meant to keep.

Clearing the collections through their own endpoints before importing. It gets the same result, but it takes several calls, leaves the channel empty in between, and requires the caller to know which collections the archive is about to bring.

Fedir-Yatsenko:
## Applicable issues

- closes #651
- closes #652
- closes #653

## Description of changes

Three gaps in the Grade C admin surface, all in the same service pair.

### Filter values (#651)

`DiscoveryDatasetStats` gains `by_agency`, so the Admin UI can render an agency picker where it has a free-text box today. `get_stats` groups by the generated `agency_key` alongside the two statuses and reports one stored spelling per key, so every name in the map is a value the `agency` filter matches - records differing only in case or spacing collapse into one entry rather than two keys standing for the same query. The map is ordered by count, largest first, ties broken by name, so the order does not depend on the query plan. `statgpt-cli channel status` renders the breakdown next to the validation and indexing ones.

<details>

<summary>CLI channel status with the agency breakdown</summary>

```
Discovery Datasets (Grade C)
  Total: 53
  Validation: VALID: 50   INVALID: 3   NOT_VALIDATED: 0
  Indexing:   INDEXED: 50   NEW: 3   OUTDATED: 0   FAILED: 0
  Agency                 Records
  Bank Indonesia (BI)         40
  OECD                        10
  IMF                          3
  Run: discovery reindex -c stat-gpt
```

</details>

### CSV export (#652)

New `GET /{channel_id}/discovery-datasets/export` returns the channel's records as a `text/csv` attachment, accepting the list endpoint's `validation_status`, `indexing_status` and `agency` filters with the same semantics. Synchronous: one channel's records are small enough to return in the response, so there is no job to poll. The file has to be the archive's `discovery_datasets.csv` byte for byte, since it is fed straight into `POST /{channel_id}/discovery-datasets/upload` on another channel - so `csv_from_dict_list` is extracted from `write_csv_from_dict_list` and both outputs go through it, with a test asserting the two agree. Columns are named from `DiscoveryDatasetBase` rather than taken from the first row, so a filter matching nothing returns a header on its own instead of failing.

### Import modes (#653)

`POST /channels/import` gains `discovery_datasets_mode` and `glossary_terms_mode`, each taking `upsert|replace` and each defaulting to `upsert`, so current behavior is unchanged. The two collections are curated independently - a channel's glossary is often maintained locally while its discovery datasets come from an archive - so each gets its own mode rather than one switch forcing a choice between leaving stale records behind and deleting terms the admin meant to keep. `replace` makes its collection match the archive, including when the archive carries no file for it at all; both modes only apply when merging into an existing channel, since with `clean_up=true`, or when the import creates the channel, there is nothing to delete. Discovery reuses the `delete_absent` reconciliation the `replace` upload already drives; glossary gains the equivalent stale delete in `_merge_terms`. `DiscoveryUploadMode` is renamed `RecordUploadMode` and shared by both endpoints - one concept, one type; the wire values `upsert`/`replace` are unchanged, so only the OpenAPI component name moves. `statgpt-cli channel import` exposes the pair as `--discovery-datasets-mode` and `--glossary-terms-mode`.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Fedir-Yatsenko:
/deploy-review

>GitHub actions run: [34368351629](https://github.com/epam/ai-dial-ci/actions/runs/34368351629)

| [deploy-review](https://chat-statgpt-backend-pr-657.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/34368351629/job/102522858019) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/34368351629/job/102526697282) |

kryachkow (approved):

PR #660, opened 2026-09-10
feat: make the data explorer link configurable per output surface #634

### LINKED ISSUE #634: Successful data queries still link the user back to the registry
### StatGPT Backend version

0.0.0

### Short description

When a data query succeeds and the data is delivered in the chat, the response still carries outbound links back to the source registry — the data explorer deep link, the dataset page, or both. The user is sent out of the platform to fetch something the platform just gave them.  Links back to the source belong to the paths where StatGPT could *not* deliver: the query failed, the query returned no rows, or the only thing worth showing is a discovery (Grade C) dataset that carries no data at all. When rows come back, the answer should stand on its own.

1. Open a channel whose data source sets `data_explorer_url`.
2. Ask any question that the data query tool answers successfully, for example
   `GDP growth 2015-2023` for a country the channel covers.
3. Read the assistant's final message and the `Query_Data` stage.

The data arrives as the table, CSV, Plotly, and Python attachments, with a short grounded summary.
No link invites the user to go and look at the same numbers in the registry.

### What do you see instead?

The response, and the tool output the agent reads before writing it, both carry explorer/dataset
links. There are three separate sources, and a single turn can hit all three:

1. A **"View data in explorer"** deep link under the execution result, pointing at the exact query
   that was just executed and rendered.
2. A **citation whose URL has been rewritten** to the registry's dataset page instead of the
   configured citation URL.
3. A **discovery datasets block** appended after the results, whose items are rendered from
   channel templates that typically include the dataset's `url`.

The agent then tends to lift these links into its own final message, because they arrive inside
the tool response it is reasoning over — not only inside the stage.

### 1. The explorer link is unconditional

`statgpt/app/utils/formatters/dataset_query.py:123-125`, in `_format_execution_result`:

```python
if data_response.url_query:
    lines.append("")
    lines.append(f'[🔍 {self._("View data in explorer")}]({data_response.url_query})')
```

This sits *after* the if/elif chain that decides whether the query succeeded, returned no data,
failed at the source, or failed to parse. Every branch falls through to it, so success is treated
exactly like failure.

The result reaches both audiences at `statgpt/app/chains/data_query/query_builder/query/execute_query.py:69-83`:
`target.append_content(response_content)` writes the user-facing stage, and the same string becomes
`DataQueryParameters.RESPONSE_FIELD` — the tool response the Supreme Agent reads. Suppressing it
in the prompt is therefore unreliable: the markdown is already in the agent's context, ready to be
copied.

### 2. The citation URL is replaced by the explorer/dataset URL

`dataset_query.py:148-153` renders the citation through `SimpleDatasetFormatter`, which at
`statgpt/app/utils/formatters/dataset_simple.py:45-53` passes `url_override_value=dataset.dataset_url`.
For SDMX v2.1 datasets, `statgpt/common/data/sdmx/v21/dataset.py:478-481` returns
`build_data_explorer_dataset_url(...)` whenever the data source sets
`use_data_explorer_for_dataset_url` (`statgpt/common/data/sdmx/common/config.py:184`). So the
citation — the one link that legitimately belongs next to delivered data — is turned into a second
registry link rather than the publisher's own reference.

### 3. The discovery block does not know whether data was found

`statgpt/app/chains/data_query/data_query_tool.py:52` runs the discovery lookup beside the
pipeline, and lines 69-76 append its block to the response whenever one was rendered:

```python
discovery_block = discovery.rendered if discovery is not None else None
if discovery_block and ChainParameters.get_invocation_source(inputs) is InvocationSource.AGENT:
    response_str = f"{response_str}\n\n{discovery_block}"
```

Neither this check nor `DiscoveryDatasetsRunner.run`
(`statgpt/app/chains/discovery_datasets/runner.py:60-103`) consults the query outcome, even though
`ExecuteQueryChain._execution_status` (`execute_query.py:88-106`) has already classified it as
`DATA_AVAILABLE`, `EXECUTED_NO_DATA`, or `FAILED`. Discovery datasets are by definition datasets
StatGPT cannot query, so their whole purpose is to offer somewhere else to look — which is exactly
what should *not* follow a successful answer. `DiscoveryDocumentMetadata.url`
(`statgpt/common/schemas/discovery_document.py:68`) is exposed as a template placeholder, so those
items are typically rendered as links.

### Why configuration cannot solve this today

The only available lever is unsetting `data_explorer_url` on the data source, which removes the
link from the no-data and failure cases too — the cases where it is genuinely useful. There is no
switch that says "link out only when we could not deliver".

## Suggested fix

**Minimum — make the explorer link conditional.** Move the `url_query` block inside the
non-success branches of `_format_execution_result` (`dataset_query.py:87-127`), so it accompanies
"no data for this period", "the request to the data source failed", and the parsing-failure note,
but not "Data received, contains N series". Add a `DatasetQueryFormatterConfig` flag
(e.g. `include_explorer_link_on_success`, default `False`) so a deployment that wants the old
behavior can keep it.

**Gate the discovery block on the query outcome.** In `data_query_tool.py`, append the block only
when the status is not `DataQueryStatus.DATA_AVAILABLE`. Better still, skip the lookup itself in
that case: the judge call is an LLM round trip whose output is then discarded. This makes the
feature match its stated intent — Grade C datasets are a fallback for what StatGPT cannot answer,
not a footnote to what it just did. Expose it as a config field on `DiscoveryDatasetsDetails` if
some channels want the block regardless.

**Keep the citation a citation.** When a query returned data, prefer the configured citation URL
over `dataset_url`, or stop applying `use_data_explorer_for_dataset_url` on the
`DatasetQueryFormatter` path specifically. The Available Datasets tool is a different case: there,
pointing at the dataset page is reasonable, since nothing was retrieved.

## Additional information

- This is not a prompting problem, and a system-prompt instruction is the wrong fix. The links are
  rendered into the tool response before the agent runs; asking the model to ignore markdown that
  is sitting in its context will hold most of the time and fail visibly the rest.
- The user-facing stage and the agent-facing tool response are built from the same string in
  `execute_query.py`, so any fix applies to both at once — no separate change is needed to stop the
  link from reaching the chat.
- Scope check when fixing: `dataset_query.py` is also reached from
  `statgpt/app/utils/formatters/dataset_availablity_query.py:82-85` and from
  `SummarizeQueriesChain._format_queries`
  (`statgpt/app/chains/data_query/query_builder/query/summarize_query.py:44-61`), which runs before
  execution and passes data responses through the same formatter.
- The MCP/widget path is worth confirming separately: `dataset_url` is carried on the query schema
  (`statgpt/common/schemas/query.py:39`) and reaches the UI widget independently of the markdown
  discussed here. Suppressing the markdown link does not suppress that field, and it may be the
  right place for the widget to keep a "view source" affordance.

Fedir-Yatsenko:
## Applicable issues

- closes #634

## Description of changes

When a data query succeeded, the response still carried a `[🔍 View data in explorer]` deep link back to the source registry — sending the user out of the platform to fetch numbers StatGPT had just delivered. The link was emitted unconditionally, so success was treated exactly like failure. Worse, the same string becomes the tool response the Supreme Agent reasons over, so the agent tended to lift the link into its own final message; suppressing it by prompt was never going to be reliable against markdown already sitting in the model's context.

The only lever available was `viewInDataExplorer` on the data source, which removes the link everywhere — including the no-data and failure paths, where linking out is genuinely the most useful thing the response can do.

**A policy per output surface.** A new `explorerLink` block on the data query tool's `details` sets, independently for each surface, when the link is rendered:

```yaml
      explorerLink:
        stage: always              # the user-facing DIAL stage
        agent: only_when_no_data   # the tool response the Supreme Agent reads
        mcp: never                 # the tool response returned to an MCP client
```

Each takes `always`, `only_when_no_data` or `never`. All three default to `always`, so stored channel configs keep today's behavior and no migration is needed. The agent/MCP split reuses the existing `InvocationSource` mechanism that already picks the `*McpOnly` message variants.

**Rendering the two audiences apart.** `ExecuteQueryChain` built one string and sent it both to the stage and to the tool response, so the two could not differ. It now resolves the stage policy and the tool-response policy separately and renders once per distinct policy — which stays a single pass in the default config, and in any config where both surfaces agree. The second pass is pure string building: the one `await` inside `format_queries` just parses a configured citation date.

`only_when_no_data` is evaluated per `DataResponse` rather than per run, so in a multi-dataset fan-out the dataset that returned rows loses its link while an empty one keeps it.

## Checklist

### DIFF
--- a/configurations/clients/sample/tools.yaml
+++ b/configurations/clients/sample/tools.yaml
@@ -126,6 +126,14 @@ tools:
           enabledStr: "False"
         markdownTable:
           enabledStr: "True"
+      # When each surface carries the "View data in explorer" deep link back to the source
+      # registry. One of: always | only_when_no_data | never. All default to `always`.
+      # Set a surface to `only_when_no_data` to keep the link on the paths where StatGPT could
+      # not deliver (no rows, or the query failed) and drop it from successful answers.
+      explorerLink:
+        stage: always
+        agent: always
+        mcp: always
       stagesConfig:
         debugOnly: true
         toolCallName: "Searching for data: {query}"
--- a/statgpt/app/chains/data_query/query_builder/query/execute_query.py
+++ b/statgpt/app/chains/data_query/query_builder/query/execute_query.py
@@ -19,8 +19,8 @@ from statgpt.app.utils.formatters import DatasetQueryFormatter, DatasetQueryForm
 from statgpt.common.auth.auth_context import AuthContext
 from statgpt.common.data.base import DataResponse, DataSetQuery
 from statgpt.common.schemas import StagesConfig
-from statgpt.common.schemas.data_query_tool import DataQueryMessages
-from statgpt.common.schemas.enums import DataParsingStatus, DataRequestStatus
+from statgpt.common.schemas.data_query_tool import DataQueryExplorerLink, DataQueryMessages
+from statgpt.common.schemas.enums import DataParsingStatus, DataRequestStatus, ExplorerLinkPolicy
 from statgpt.common.schemas.tool_details import StageDescriptor
 
 from .summarize_query import SummarizeQueriesChain
@@ -32,11 +32,13 @@ class ExecuteQueryChain:
         stages_config: StagesConfig,
         stage: StageDescriptor,
         messages: DataQueryMessages,
+        explorer_link: DataQueryExplorerLink,
         summarize_queries_chain: SummarizeQueriesChain,
     ):
         self._stages_config = stages_config
         self._stage = stage
         self._messages = messages
+        self._explorer_link = explorer_link
         self._summarize_queries_chain = summarize_queries_chain
 
     async def summarize_dataset_queries(self, inputs: dict) -> dict:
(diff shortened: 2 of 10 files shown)


Fedir-Yatsenko:
/deploy-review

>GitHub actions run: [34473499428](https://github.com/epam/ai-dial-ci/actions/runs/34473499428)

| [deploy-review](https://chat-statgpt-backend-pr-660.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/34473499428/job/102858665315) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/34473499428/job/102861567605) |

kryachkow (approved):

PR #671, opened 2026-09-15
feat: make the data presentation prompt section configurable per channel #670

### LINKED ISSUE #670: Make the "Data Presentation to User" prompt section configurable per channel
## StatGPT Backend version

latest

## What is the problem this feature will solve?

The supreme agent's system prompt already exposes several sections as per-channel overrides — `generalSection`, `toolUsageSection`, `noCalculationsSection`, `userUiContextSection` — so a deployment can tailor them to its own data and tone without forking the prompt.

The **Data Presentation to User** section is the exception: it is hardcoded in `statgpt/app/default_prompts/assets/supreme_agent.yaml:85-107`. It is also one of the most deployment-specific parts of the prompt — how sources are cited, the retrieval-vs-coverage wording, how the factual time period and the default time filter are reported, and how forecasts are flagged all depend on the channel's data and on what the surrounding UI already shows the user. Today the only way to change any of it is to edit the shipped asset, which affects every channel and is lost on upgrade.

## What is the proposed feature or solution?

Extract the section into a configurable section, following the existing pattern exactly:

- Move the current content from the `systemPrompt` body into a `defaultDataPresentationSection` entry in `supreme_agent.yaml`, leaving a `{data_presentation_section}` placeholder under the `## Data Presentation to User` heading.
- Add `default_data_presentation_section` to `SupremeAgentDefaultPrompts`.
- Add `data_presentation_section: str = ""` to `SupremeAgentConfig` (`statgpt/common/schemas/channel.py`), with the same "if empty, the default content is used" semantics as its siblings.
- Resolve it in `SupremeAgentChain._create_prompt_template` as `channel_config.supreme_agent.data_presentation_section or supreme_agent_default_prompts.default_data_presentation_section`.

Default behavior is unchanged for every existing channel — the rendered prompt is byte-identical unless a channel sets an override.

## What alternatives have you considered?

- **Leaving it in `additionalContext`** — rejected; additional context is explicitly lower priority than the main instructions, so it cannot replace presentation rules, only append to them.
- **Editing the shipped asset per deployment** — rejected; it is not per-channel and does not survive an upgrade.

Fedir-Yatsenko:
## Applicable issues

- closes #670

## Description of changes

Extracts the hardcoded **Data Presentation to User** section of the supreme agent's system prompt into a configurable section, so a channel can tailor it the way it already tailors `generalSection`, `toolUsageSection`, `noCalculationsSection` and `userUiContextSection`.

The section's content moves from the `systemPrompt` body into `defaultDataPresentationSection` in `supreme_agent.yaml`, leaving a `{data_presentation_section}` placeholder under the heading. `SupremeAgentConfig` gains `dataPresentationSection`, resolved in `SupremeAgentChain._create_system_prompt` with the same "if empty, the default is used" semantics as its siblings.

Unlike its siblings, the section is inlined into the template string rather than passed to `.partial(...)`: `str.format` is single-pass, so placeholders the section carries would otherwise render literally — starting with the `{today_date}` of the default content, in the forecast rule. Inlining keeps them resolving, so an override may reference any of the prompt's placeholders (`{today_date}`, `{chat_bot_name}`, `{chat_bot_domain}`, …), not just the one the default uses. The trade-off is that a stray `{` in an override now raises at format time instead of passing through.

Default behavior is unchanged: the rendered default prompt was diffed against `development` and is byte-identical. The existing prompt unit tests were extended to pin both the placeholder resolution and an override fully replacing the default.

## Checklist


statgpt/common/schemas/channel.py
```diff
@@ -81,6 +81,14 @@ class SupremeAgentConfig(BaseYamlModel):
             " If empty, the default content is used."
         ),
     )
+    data_presentation_section: str = Field(
+        default="",
+        description=(
+            "Custom content for the 'Data Presentation to User' section of the system prompt."
+            " May reference the {today_date} placeholder."
```
navalnica: it may also reference other placeholders, not only today date

navalnica (dismissed):

statgpt/common/schemas/channel.py
navalnica: I'd update the field description

Fedir-Yatsenko: Fixed

navalnica (approved):

